Issue #30, opened 2020-10-28
env file


enarankhuu:
Which text insert into .env.local? I dunno

techrangle:
Add these lines but i am trying to find store front api url and store front api token.

```
BIGCOMMERCE_STOREFRONT_API_URL=
BIGCOMMERCE_STOREFRONT_API_TOKEN=
BIGCOMMERCE_STORE_API_URL=
BIGCOMMERCE_STORE_API_TOKEN=
BIGCOMMERCE_STORE_API_CLIENT_ID= 
```

techrangle:
save this in .env.local file on the root project.

FredvanRijswijk:
If you use the vercel cli you can use this `vercel env pull .env.local` super easy and the right keys at once

Strutsagget:
Hi
Where can I find these settings?

BIGCOMMERCE_STOREFRONT_API_URL=
BIGCOMMERCE_STOREFRONT_API_TOKEN=
BIGCOMMERCE_STORE_API_URL=
BIGCOMMERCE_STORE_API_TOKEN=
BIGCOMMERCE_STORE_API_CLIENT_ID= 

After creating an API user I get a file with:
ACCESS TOKEN:
CLIENT ID: 
CLIENT SECRET: 
NAME: 
API PATH: 

adding the API path as both URLs, token as both tokens and id as id does not seem to work as the application crashes on launch without being able to read the message.
TypeError: Cannot read property 'message' of undefined in the store front api hooks.

Rando1993:
https:>UI_+_bot+link_//api_run

techrangle:
> Hi
> Where can I find these settings?
> 
> BIGCOMMERCE_STOREFRONT_API_URL=
> BIGCOMMERCE_STOREFRONT_API_TOKEN=
> BIGCOMMERCE_STORE_API_URL=
> BIGCOMMERCE_STORE_API_TOKEN=
> BIGCOMMERCE_STORE_API_CLIENT_ID=
> 
> After creating an API user I get a file with:
> ACCESS TOKEN:
> CLIENT ID:
> CLIENT SECRET:
> NAME:
> API PATH:
> 
> adding the API path as both URLs, token as both tokens and id as id does not seem to work as the application crashes on launch without being able to read the message.
> TypeError: Cannot read property 'message' of undefined in the store front api hooks.

[...]

Issue #29, opened 2020-10-28
[ISSUE] can't add products to cart


buty619:
## Description

* On my local all actions related to mybigcommerce (for example add to cart) always return Error `[BigcommerceApiError]: Big Commerce API error (404)`, maybe im missing some configuration?

## Img

<img width="1245" alt="Screen Shot 2020-10-27 at 10 08 09 PM" src="https://user-images.githubusercontent.com/36422189/97386017-f047bf80-18a0-11eb-9fb3-24372666904c.png">

enarankhuu:
Hey, I wanna show your .env.local file. I dunno which env file to use!!

okbel:
.env.template was added! Keep in mind that this is tightly coupled with BigCommerce so you'll need env vars to run it locally.

buty619:
hello @okbel  thanks for all you help but im still having same issue i already create the .ev.local file whit the key get from versel, there is something else im missing? maybe local env should be different to the once generate on versel?

<img width="1263" alt="Screen Shot 2020-10-28 at 5 13 42 PM" src="https://user-images.githubusercontent.com/36422189/97502588-6a2f8580-1941-11eb-8488-463e48147065.png">

<img width="339" alt="Screen Shot 2020-10-28 at 5 14 17 PM" src="https://user-images.githubusercontent.com/36422189/97502598-6ef43980-1941-11eb-9f6e-1ff45dca704f.png">

Abdulsalam0001:
I am  having the same issue...

Julianwan:
Has anyone found a solve for this? Just curious... I'm having a similar set of issues

Issue #26, opened 2020-10-28
[BUG] log in and Sign Up always return error


buty619:
### For Sign Up
* on local and in the deployed page when I try to make a sing up always get `An unexpected error ocurred` and the logs says `Error: Internal server error at FetcherError.CommerceError [as constructor]` but then in the `mybigcommerce` i can see the user created as expected

### For Log In

* on local and in the deployed page  always return `An unexpected error ocurred. Did you forgot your password?` and the logs says `Error: Internal server error at FetcherError.CommerceError [as constructor]`

okbel:
Did you set the env vars right? We just merged .env.template. LMK!

buty619:
yes i set it whit the vercel shell and making `vercel env pull` but is set as usual .env, so i don't have the error asking for the env vars on my local, also @okbel  thanks for answering ¡¡

robbieaverill:
I just noticed this on demo.vercel.store too - is the login API supposed to return a 500 when I put an invalid username/password in? Seems like a 40x would be better.

okbel:
Bringing BigCommerce to debug this @RayeEThompson

Balance8:
I can corroborate that I am experiencing the same issue, env variables pulled via vercel CLI, account created according to the regex.  ```Error: Internal server error
    at FetcherError.CommerceError [as constructor]```
User is generated in big commerce 

![image](https://user-images.githubusercontent.com/36809369/98073351-031c4000-1e1d-11eb-942f-63aeb68fa3c7.png)

magicspon:
Snap... I'm also seeng this error.

RayeEThompson:
Hey guys, working with the team to debug this. We have the issue reported to BigCommerce engineering. I will follow up as soon as possible.

sadpvndv:
I have this error on SignUp and Login `Error: Invalid credentials at FetcherError.CommerceError [as constructor]` on local and in the deployed page

murillo94:
Same error here, tried to login in and sign up takes the error:

|view|url|response|
|-----|--|---------|
[...]

Issue #24, opened 2020-10-27
Sign-up is not working


gr-qft:
Whatever email and password given when signing-up, there's always the error message about passwords must be greater than 7 characters with at least a numeric character. It seems to me authentication is not fully implemented yet?

okbel:
We just added an .env.template. Have you tried that? LMK!

gr-qft:
@okbel I still have not been able to sign up, still seeing the same message (my password surely is longer than 7 chars and includes numbers).

frannunezrivera:
I'm having this problem too:
@gr-qft  if you check in the bigcommerce - customers panel you will see your user has been created. I believe the problem is when logging in. (in this case probable after user is being created)

@okbel The sign-in hook is returning invalid credentials (error 400 : bad request) but the API request is returning an error 500. POST http://localhost:3000/api/bigcommerce/customers/login 500 (Internal Server Error). Not sure why because it looks the params are being passed correctly... I didn't have time to check the @bigcommerce npm package yet.

okbel:
@gr-qft What data are you giving to sign up?
And yes, auth has been implemented. Try https://demo.vercel.store to check it out live.

Yeah, @frannunezrivera it's using those hooks so I think it may be masking some errors with the password lengths and criteria

gr-qft:
@okbel The problem is not with the demo site. The problem occurred when I deployed vercel commerce repo on my vercel account. I still have a problem, but the deployed repo is 8d old, and I'm running into some deployment problem (hook, module etc.) so I cannot test it now.

iamfeek:
mm yeah. I am facing the same problem as @gr-qft. 
Repo deployed 2 days ago, testing on local. 
.env areismatching that of Vercel's env vars.

stevencmh:
> @okbel The problem is not with the demo site. The problem occurred when I deployed vercel commerce repo on my vercel account. I still have a problem, but the deployed repo is 8d old, and I'm running into some deployment problem (hook, module etc.) so I cannot test it now.

Hi, did you get this working? I'm seeing the same thing on my deployment.

bobbiejaxn:
> > @okbel The problem is not with the demo site. The problem occurred when I deployed vercel commerce repo on my vercel account. I still have a problem, but the deployed repo is 8d old, and I'm running into some deployment problem (hook, module etc.) so I cannot test it now.
> 
> Hi, did you get this working? I'm seeing the same thing on my deployment.

[...]

PR #31, opened 2020-10-28
Add .env.template

marbiano:
This PR adds an empty `.env.template` file for people to use as a starter. It also adds notes in the README in order to be able to use it properly.

Fixes https://github.com/vercel/commerce/issues/30, and potentially helps on https://github.com/vercel/commerce/issues/29, https://github.com/vercel/commerce/issues/26, https://github.com/vercel/commerce/issues/24, https://github.com/vercel/commerce/issues/17 and https://github.com/vercel/commerce/issues/14.


FredvanRijswijk (commented):
You also need to add `vercel env pull .env.local` because thats the easiest way to add the right keys at once to a local env file

okbel (approved):

marbiano:
@FredvanRijswijk thanks for the review! I think that could be nice to add to the README, but with some kind of indication that it only applies if you are using Vercel and have an app already set on it. For any other use case, manually setting the `.env.local` file is the way to go.

PR #33, opened 2020-10-28
Cart loading state

marbiano:
This PR adds a basic loading state so the screen doesn't get an unnecessary flash of empty state.

In a subsequent iteration, it would be great to integrate this loading state with the actual cart layout and make it a pulse skeleton using Tailwind. But I wanted to start with something that fixes the visual issue.


marbiano:
@okbel This is actually not working appropriately because of https://github.com/vercel/commerce/issues/4, when there are no items in the cart the API should return an empty array, not undefined. That would allow us to use the way that SWR indicates to distinguish the loading scenario (`data == null`).

okbel:
Thanks, @marbiano. Could you use the skeleton component?

marbiano:
@okbel yup, I can take care of that. Do you happen to know if the BigCommerce hook to fetch cart is gonna be fixed sometime soon? Otherwise, there's no way for me to figure out if cart content is loading or just empty with the current hooks.

okbel:
What do you mean? Fetching on load means loading true until data is present (not null), right?

marbiano:
From what I understand of how SWR works, `data` starts as null and that's how you know that it's loading, until it's not null anymore, which means that data is loaded.

This is from their docs:

```
function Profile() {
  const { data, error } = useSWR('/api/user', fetcher)
  if (error) return <div>failed to load</div>
  if (!data) return <div>loading...</div>
  return <div>hello {data.name}!</div>
}
```

So data can't return null as it's happening right now when there are no items on the cart, because then I have no way of distinguishing between a loading cart and an empty cart.

`react-query` has a primitive for this (isLoading) that I don't see it present on SWR. Do you know of any way to do this that doesn't imply for BigCommerce folks to fix the hook?

lfades:
@marbiano `data` has 3 values, and the cart hook also has a `isEmpty` prop.

- `isEmpty: true` -> the cart loaded and it's empty
- `data: undefined` -> the cart is loading
- `data: null` -> the cart loaded and there's no cart (and `isEmpty` is set to `true`)
- `data: Data` -> the cart loaded and it has items

marbiano:
@lfades really good to know, thank you! I'm proficient with `react-query` but it's my first time using `SWR`. 😅

I'll take care of this stuff tonight!

lfades:
@marbiano Let me know when it's ready for review! 🙏 🙇

marbiano:
@lfades this is now ready for review!

Here's a quick gif of the loading skeleton:

movie8

components/cart/CartEmpty/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { Bag } from '@components/icons'
+import { Button } from '@components/ui'
+
+const CartEmpty: React.FC = () => {
```
lfades, line 4: Make this an import, it's valid but looks wrong:

```
import { FC } from 'react'
```

components/cart/CartEmpty/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <span className="border border-dashed border-secondary rounded-full flex items-center justify-center w-16 h-16 bg-primary p-12 rounded-lg text-primary">
+        <Bag className="absolute" />
+      </span>
+      <h2 className="pt-6 text-2xl font-bold tracking-wide text-center">
+        Your cart is empty
+      </h2>
+      <p className="text-accents-6 px-10 text-center pt-2">
+        Biscuit oat cake wafer icing ice cream tiramisu pudding cupcake.
+      </p>
+      <Button href="/" Component="a" className="block mt-8">
```
lfades, line 16: Should be using `next/link` too.

components/cart/CartSkeleton/index.tsx
```diff
@@ -0,0 +1,76 @@
+const CartSkeleton: React.FC = () => {
```
lfades, line 1: Import react before using the `FC` type.

lib/hooks/use-cart.ts
```diff
@@ -0,0 +1,30 @@
...
+  return {
+    items: data?.line_items.physical_items ?? [],
+    isLoading: data === undefined,
+    isError: error,
+    isEmpty: isEmpty && data === null,
+    subtotal,
+    total,
+    currency: data?.currency,
+  }
+}
```
lfades, line 30: Move this to the page, we don't want to create a hook of the hook unless it's strictly required due to something that is app specific, is okay if it's more code.

pages/cart.tsx
```diff
@@ -21,123 +22,88 @@ export async function getStaticProps({
...
-              <Text>
+    <div className="px-4 pt-2 sm:px-6 flex-1">
+      <Text variant="pageHeading">My Cart</Text>
+      {isLoading ? (
+        <CartSkeleton />
+      ) : isError ? (
+        <div className="mt-2">Failed to load</div>
+      ) : isEmpty ? (
+        <CartEmpty />
+      ) : (
```
lfades, line 44: If you handle `isEmpty` first, you don't have to check for `data === undefined`, only for `!data`. Also handle the error first (`data` is a getter)

lib/hooks/use-cart.ts
marbiano: The reason I put this into a custom hook is that I think it opens the door to start thinking about multiple providers and alternate ways of fetching data, since the provider logic and the fetching mechanism would be encapsulated in here. 

Thoughts?

pages/cart.tsx
marbiano: I don't follow what you mean @lfades, care to expand? Thanks!

marbiano: @lfades by the way, `isEmpty` starts being true, then it goest to false while loading, and then it comes back to the proper value after fetch is done. So I don't think I can fully rely on it to determine if the cart is loading.

lib/hooks/use-cart.ts
lfades: @marbiano Each provider will have its own hooks ideally, like with BC, they should be used in the same way but the structure of the data is different, to make it work for multiple providers you would need to map the entire data into a completely different object because they don't follow a shared schema.

I think this is nice but not for now, again unless we really need it. The example should showcase how some external hooks for data fetching + UI components can create a powerful Commerce example. I don't want to start creating abstractions so soon. (We can always PR to the BC hooks if something needs to be done in a hook).

pages/cart.tsx
lfades: This: https://github.com/vercel/commerce/pull/33#issuecomment-718966924 <- The order I used there actually makes a difference.

lib/hooks/use-cart.ts
marbiano: @lfades makes sense, I'll put it back into the component.

pages/cart.tsx
marbiano: Screen Shot 2020-10-30 at 10 41 36 PM

That's what happens when I load `/cart/` directly. If I follow the order you proposed, I get a flash of empty cart no matter what because of `isEmpty` being initially true.

marbiano: I just put the logic back into the page component and moved error to the top of the rendering logic, but I still keep a custom `isLoading` flag because I don't think `isEmpty` is usable on the initial render. Please take a look at the update.

lfades: @marbiano You're right, there must be something we can do in the hook itself then 🤔

marbiano: @lfades maybe we can live with this solution until the hook is updated by the BC team?

lfades: I can make the update (I did those hooks, we moved them to a BC repo before releasing the example), basically we want a `isLoading` prop alongside the `isEmpty` one? makes sense 🤔

marbiano: Oh I didn't know that! Yeah, `isLoading` would be perfect and it would also mean almost no update on a potential `react-query` alternate implementation. 😁

pages/cart.tsx
```diff
@@ -35,109 +39,78 @@ export default function Cart() {
...
-            <div className="my-6">
-              <Text>
+    <div className="px-4 pt-2 sm:px-6 flex-1">
+      <Text variant="pageHeading">My Cart</Text>
+      {error ? (
+        <div className="mt-2">Failed to load</div>
+      ) : isLoading ? (
+        <CartSkeleton />
+      ) : items.length > 0 ? (
+        <div className="grid lg:grid-cols-12">
```
franklinjavier, line 50: What do you think of isolating this part into a new component? You can get advantages of composition for this case

```jsx
{error && <CartError />}
{isLoading && <CartSkeleton />}
{!!items.length && <Cart items={items} />}
```

okbel:
@marbiano could you solve the conficts? :)

lfades:
Closing as we now have loading state for hooks.

PR #41, opened 2020-10-29
Cart Count will be Number of items rather than sum total quantity

### LINKED ISSUE #21: [BUG] - Total of the Cart not corresponds with the number of items
In the cart section, the number of items doesn't alter the total sum of the purchase.
Screen Shot 2020-10-27 at 18 14 10

Awesome work guys, I can take this issue if you want 😄

pruthvi2103:
Fixes #21  earlier cart used to have sum of the total items rather than no. of items

### DIFF
--- a/components/core/UserNav/UserNav.tsx
+++ b/components/core/UserNav/UserNav.tsx
@@ -14,7 +14,7 @@ interface Props {
   className?: string
 }
 
-const countItem = (count: number, item: any) => count + item.quantity
+const countItem = (count: number, item: any) => count + (item.quantity > 0 ? 1 : 0)
 const countItems = (count: number, items: any[]) =>
   items.reduce(countItem, count)
 


laithyounes (approved):

laithyounes (approved):

components/core/UserNav/UserNav.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
   className?: string
 }
 
-const countItem = (count: number, item: any) => count + item.quantity
+const countItem = (count: number, item: any) => count + (item.quantity > 0 ? 1 : 0)
```
KyleBurger, line 17: This ternary could instead just be: 
`const countItem = (count: number, item: any) => count + (item.quantity > 0)`

pruthvi2103: Hey, the above fix by you sounds just as good if we were using JS since it is a TypeScript project I wanted to make sure that I ensured the type-safety Using your suggested code leads to an error occurring that states: 'Operator + cannot be applied to types 'number' and 'boolean' as shown in the image below

Screen Shot 2020-11-02 at 4 36 58 AM

Since the change is minor I do not want to go around messing with the Ts settings just so I don't get this error. I am fairly new to React (Learnt React 4 months ago) so I may be in the wrong too.

KyleBurger: Oooo, right good call 👍. I suppose this is more readable anyway.

KyleBurger (approved):

Issue #44, opened 2020-10-29
Bug: PDP page doesn't format prices correctly


robbieaverill:
See https://demo.vercel.store/product/next-js-short-sleeve-unisex-t-shirt

The price is 13.50 USD, but it's rendered as 13.5 USD. Should be 13.50 USD.

![image](https://user-images.githubusercontent.com/5170590/97606506-18ccd800-19cd-11eb-95a6-f9f701449f9c.png)

bntzio:
Fixed on #47

robbieaverill:
Nice, thanks @bntzio

PR #47, opened 2020-10-29
Properly format items pricing

bntzio:
This uses the price hook in the slug pages to properly format the item price in the individual product pages.

See #44


lfades (approved):
Thank you!

PR #54, opened 2020-10-31
Fix: hasScrolled functionality

matiasperz:
The function inside debounce() was never executed because debounce returns a new function.


okbel:
Good catch!

momoiicom:
Completed by Sandcastle

PR #57, opened 2020-11-01
Update core/Layout function call

eliamartani:
## Description

A piece of code was repeating a function call in a manner it could be easily fixed by just applying DRY principle. Fortunately I only found this one.

Please see file changes. Hope this helps.


okbel (approved):
NICE!

PR #73, opened 2020-11-06
Fix responsiviness on search page

matheuskuster:
This PR fixes responsiveness on the search page. It hides the Sort and Filter options and adds a button to show them on mobile devices. I hope you guys like it, any suggestions for changes in this behavior are welcome.

## Before
image

## After
<div display="flex">

</div>


pages/search.tsx
```diff
@@ -66,101 +69,155 @@ export default function Search({
...
+
+                <button
+                  className="md:invisible text-accents-8 outline-none"
+                  onClick={() => setShowMobileMenu(prev => !prev)}
+                >
+                  Filter & Sort
+                </button>
+              </div>
+
+              {showMobileMenu && (
```
omurilo, line 213: Hi there, my 2 cents is, why you choose use `showMobileMenu = true` to show mobile content and not use `showMobileMenu = false` to show desktop content? I see a duplicated `<Categories /> <Designers /> and <Sort />` elements on the final markup.

pages/search.tsx
```diff
@@ -66,101 +69,155 @@ export default function Search({
...
+                  </div>
+                ) : (
+                  <>Searching...</>
+                )}
+
+                <button
+                  className="md:invisible text-accents-8 outline-none"
+                  onClick={() => setShowMobileMenu(prev => !prev)}
+                >
+                  Filter & Sort
```
omurilo, line 209: I think the better choose for button is an filter icon.

matheuskuster: Me too, I didn't apply on first hand because I was running out of time. Now it's an icon. Updated PR description image too.

pages/search.tsx
matheuskuster: Actually, there are two "menus" now, one appears on `md+` devices and another on `md-` devices. This behavior is controlled by the tailwind classes, that's why the duplicated `<Categories /> <Designers /> and <Sort />`.

The `showMobileMenu` state is used to control only the click on the "Filter & Sort" button.

marceloF5 (approved):

pages/search.tsx
```diff
@@ -66,101 +70,155 @@ export default function Search({
...
-                  }}
-                >
-                  <a>{node.name}</a>
-                </Link>
-              </li>
-            ))}
-          </ul>
+        <div className="col-span-2 hidden md:block">
+          <Categories />
+          <Designers />
```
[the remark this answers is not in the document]
matheuskuster: I tried doing that but `categories` and `brands` have different types, so I couldn't pass them as a unique prop. The only way is receiving the `map` as children, which leads to almost the same code repeat. I pushed the result that I had, give it a look. Maybe we can pass the List of options to a shared `component`.

pages/search.tsx
omurilo: Exactly because there are two identical menus but with different views, I said it could be refactored to use the variable when it is mobile or not, thus saving code.

okbel:
A new PR that added dropdowns to solve this was just merged. Thank you!

Issue #79, opened 2020-11-08
Server Error: Image is missing required "src" property.


leahtard:
When first running `yarn dev` after build, I was seeing the following error:

```
Server Error
Error: Image is missing required "src" property. Make sure you pass "src" in props to the `next/image` component. Received: {"width":540,"height":540,"quality":"85"}
```

This occurred because I connected to an existing BigCommerce sandbox -- as opposed to creating a new site on creation. In this sandbox, I had a few products without any images which were causing this error. It would be great if the commerce package could account for this. 

My proposal:
1. Filter out products from the Grid displays on the homepage when they do not have images.
2. Provide a placeholder/fallback image for the ProductCard.

Cheers, Leah

leahtard:
I get this a try in this PR: https://github.com/vercel/commerce/pull/80

okbel:
This is great!!! Thank you. _checking_

PR #80, opened 2020-11-08
Filter products from the homepage grid display when they do not have …

leahtard:
Potential solution to issue reported: https://github.com/vercel/commerce/issues/79
1. Filter out products from the Grid displays on the homepage when they do not have images.
2. Provide a placeholder/fallback image for the ProductCard.

I am new to TS so I am a little unsure how I have handed the filter on line 49 of index.ts. After doing some searching/reading, I though `p?.node?.images?.edges?.length` would fix the build error `Type error: Object is possibly 'null' or 'undefined'` but this didn't solve the error. Applying `p.node.images.edges!.length` ended up working but I wasn't quite sure if it was quite right. Totally open to feedback here :)


components/product/ProductCard/ProductCard.tsx
```diff
@@ -80,7 +80,7 @@ const ProductCard: FC<Props> = ({
             <div className={s.imageContainer}>
               <Image
                 quality="85"
-                src={src}
+                src={src || '/product-img-placeholder.svg'}
```
marceloF5, line 83: @leahtard 
I would like to hear your motivation to import an SVG inline instead of importing it at the top and reusing it here and on line 60. Importing directly here, you may lose more performance, nothing significant, but it would be a request for each image not found. I would assume that there would always be no image on any of the items and import the placeholder at the top only once and use that reference for those that were not found. What do you think?

leahtard: Thank you for your review @marceloF5. I pushed up a small update that fixes the repeating image references. I did try to place the `svg` markup inline, but the Image component is type checking for an image reference.

Cheers, Leah

marceloF5: Nice @leahtard! 🚀

pages/index.tsx
```diff
@@ -44,7 +44,9 @@ export async function getStaticProps({
   // We prefer to do the computation at buildtime/servertime
   const { featured, bestSelling } = (() => {
     // Create a copy of products that we can mutate
+    // Filter products that do not have images
     const products = [...newestProducts]
+      .filter((p) => p.node.images.edges!.length > 0)
```
okbel, line 49: I'd rather not add another iteration over the products array. What about rendering conditionally the image component or a placeholder. (The more computations we can avoid, the better)

okbel (changes requested):

okbel:
Inactive thread, archiving until conversation gets active again.

PR #91, opened 2020-11-18
put filters inside a dropdown in mobile devices

ricardolmsilva:
Hi everyone, this is my first pull request contributing with open source community, so forgive me if I'm not doing it in the best way.

I found an issue in small screens, when we are on the products page the product filters don't look great, so I decided to help with this. Just a note, this was also my first time using Tailwind so I hope that I have used it right.

Here is the result, I hope you like it. 

result


ricardolmsilva:
For some reason, in this preview mode, I can see a different result that I got in my local environment.
Here is what I can see in the page on the desktop and when opening the dropdown on mobile.

dropdown-and-desktop

okbel:
Looks incredible! Thank you!! Keep them coming (:

okbel (approved):

PR #103, opened 2020-11-29
Button and I18n Widget Fix

Regaron:
Fixed Button component missing onClick prop and I18n widget image being outside button also the usernav item not showing when cart is empty

### DIFF
--- a/components/common/I18nWidget/I18nWidget.tsx
+++ b/components/common/I18nWidget/I18nWidget.tsx
@@ -72,7 +72,7 @@ const I18nWidget: FC = () => {
       <div className="absolute top-0 right-0">
         {options?.length && display ? (
           <div className={s.dropdownMenu}>
-            <div className="flex flex-row justify-end px-2 mt-2">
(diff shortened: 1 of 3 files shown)


luke-denton-aligent:
@Regaron Thanks for bringing this functionality back in! I wonder if we can also update the Button component to fix the Proceed to Checkout "button"? That element is actually an `<a>`, however uses `Button.tsx` to render, which means clicking on it also does nothing

okbel:
Screen Shot 2020-12-02 at 11 18 55 AM

I can still see the issue

Regaron:
It’s showing correctly in latest preview:
https://commerce-e57urrj2r.vercel.app/

okbel:
Had to close it because it has been solved already. Thank you!!

luke-denton-aligent:
Hi @okbel, I'm curious why this was closed. If we look at the Button.tsx component in the `master` branch, the `onClick` prop still isn't handled. This means that anyone who is starting a new project from the current head of master will get buttons that don't work :confused:

PR #124, opened 2020-12-17
Fix handleScroll event handler - Navbar component

ollyswanson:
Hi, this is my first pull request to an open source project, so I'm sorry if I've done anything wrong!

I believe that the previous implementation of the scroll detection is attaching a new event handler on every render of the component. `handleScroll` has a new reference on each render and therefor `useEffect` is running on each render (in a loop due to the setting of state), as well as the call to `document.removeEventListener` being passed a function with a different reference value to the call to `document.addEventListener`.

P.S.
I've just noticed that my Tailwind plugin has reordered the class names in the JSX, I'm sorry  for the unnecessary commit lines.


okbel (approved):
Looks good! Thank you!

Issue #119, opened 2020-12-09
Uncaught (in promise) TypeError: Expected "nextInternalLocale" to be a string


stevencmh:
Hi,

Error occurs when using the All Categories and All Designers filter at the bottom left of the home page.

<img width="1183" alt="Screenshot 2020-12-09 at 10 49 35" src="https://user-images.githubusercontent.com/35059438/101620285-3f3e5400-3a0c-11eb-9dbc-c55335b45726.png">

markcmurphy:
I experienced this when searching while changing the locale to ES. This thread helped me fix the issue: https://github.com/vercel/next.js/issues/19521 and it looks like they've addressed it in a canary release: https://github.com/vercel/next.js/pull/19766

stevencmh:
Hi,

Thanks! … I'll just pulled and merged canary but it's still throwing the same error.

Just looking on the demo site, and the filter on the right-hand side also results in an error.

PR #125, opened 2020-12-20
fix locale category rewrite

renanfeluck:
The next.js is having an error while trying to automatically rewrites each source and destination prefixed to handle the configured locales because of the search category route. I added a locale false to this route to fix this problem

This is the error that was fixed when we tried to change the category: 

locale bug


renanfeluck:
This pr is related with this issue https://github.com/vercel/commerce/issues/119

okbel (approved):

Issue #111, opened 2020-12-02
Not displaying product description on page reload (demo.vercel.store website)


ulysset:
Hi,
This problem has been found when testing the website on desktop Chrome 87, Safari and another computer on a different network :
When refreshing a product page the description disappears (see image bellow)
Failed to load resource: the server responded with a status of 500 ()` 
<img width="1440" alt="Screenshot 2020-12-02 at 20 10 39" src="https://user-images.githubusercontent.com/13087992/100919750-7d061e80-34da-11eb-8163-fbe27326d724.png">

stevencmh:
Hello,

I'm seeing the same with these two warnings:

`Warning: Prop `dangerouslySetInnerHTML` did not match.`

`Warning: Invalid DOM property `shape-rendering`. Did you mean `shapeRendering`?`

PR #126, opened 2020-12-20
Fix text component

renanfeluck:
The text component was creating an <p> element and receiving an HTML inside it. But when you have a <p> tag inside another <p> tag you can have some problems.

In this case, the product description for example was not being displayed after you reload the page


renanfeluck:
This PR is related to this issue:

https://github.com/vercel/commerce/issues/111

PR #132, opened 2020-12-23
Prevent system navigation on swipe

shuding:
This PR fixes the iOS system navigation gesture (swipe to go back/forward) behavior for the Slider component. And it will no longer trigger the browser navigation when you swipe from the left/right side of your screen.

It's implemented by preventing the `touchstart` event when it's near the edges.

Reference: https://pqina.nl/blog/blocking-navigation-gestures-on-ios-13-4

### DIFF
--- a/components/product/ProductSlider/ProductSlider.tsx
+++ b/components/product/ProductSlider/ProductSlider.tsx
@@ -1,5 +1,5 @@
 import { useKeenSlider } from 'keen-slider/react'
-import React, { Children, FC, isValidElement, useState } from 'react'
+import React, { Children, FC, isValidElement, useState, useRef, useEffect } from 'react'
 import cn from 'classnames'
 
 import s from './ProductSlider.module.css'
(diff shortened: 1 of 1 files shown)


components/product/ProductSlider/ProductSlider.tsx
```diff
@@ -17,8 +18,35 @@ const ProductSlider: FC = ({ children }) => {
...
+      // if the touch area overlaps with the screen edges
+      // it's likely to trigger the navigation. We prevent the
+      // touchstart event in that case.
+      if (
+        touchXPosition - touchXRadius < 10 ||
+        touchXPosition + touchXRadius > window.innerWidth - 10
+      ) event.preventDefault()
+    }
+
+    sliderContainerRef.current!
```
montogeek, line 39: What is this syntax? `current!.`?

components/product/ProductSlider/ProductSlider.tsx
```diff
@@ -17,8 +18,25 @@ const ProductSlider: FC = ({ children }) => {
...
+
+    return () => {
+      sliderContainerRef.current!
+      .removeEventListener('touchstart', preventNavigation)
+    }
+  }, [])
+
   return (
-    <div className={s.root}>
+    <div className={s.root} ref={sliderContainerRef}>
```
Badyl1g, line 49: Hey :), asking out of curiosity: why not use the onTouchStart event directly on the div declaratively instead of attaching the listener via ref?

shuding: Haha good question. Actually there's no particular reason, I'm just used to manage listeners like that :D

okbel:
Super neat! Thank you @shuding !

okbel (approved):

Issue #136, opened 2020-12-29
Image slider shows all images when navigating from product page to product page.


tkapitein:
https://user-images.githubusercontent.com/37157784/103277396-2d2b5200-49c9-11eb-8891-a2f842ec533d.mov

Used browser: Google Chrome

When you visit a product page and visit another product page via the shoppingcart, all product images will show in the slider.  It seems that the product slider components acts on the props of the previous productpage.

okbel:
Amazing thank you

PR #137, opened 2020-12-29
Fix ProductSlider

tkapitein:
Fixed issue #136  by adding a unique key prop to the ProductSlider component.


tkapitein:
Will close this PR as it can be fixed without adding dependencies.

PR #138, opened 2020-12-29
Fix Product Image slider

tkapitein:
Fixed issue #136 by adding the product entityID as an unique key prop to the `ProductSlider` component.


okbel:
What a good catch, thank you 🙏

okbel (approved):

PR #145, opened 2021-01-06
Focus trap and Modal Functionality

okbel:
- Escape and Enter keys functionality and behaviour
- Focus trap
- New `focusFirst` prop to attempt focus the first focusable item within modal children, works with dynamic components.


components/ui/Modal/Modal.tsx
```diff
@@ -7,45 +7,52 @@ import {
...
 
-const Modal: FC<Props> = ({ children, open, onClose }) => {
+const Modal: FC<Props> = ({ children, open, onClose, onEnter = null }) => {
   const ref = useRef() as React.MutableRefObject<HTMLDivElement>
 
+  const handleKey = (e: KeyboardEvent) => {
+    if (e.key === 'Escape') {
+      return onClose()
+    }
+  }
```
lfades, line 26: Use `useCallback` here and then add `handleKey` to the deps of the `useEffect` below, that way the function `onClose` will be called properly if it's changed by the parent component for any reason.

lib/focus-trap.tsx
```diff
@@ -0,0 +1,64 @@
...
+      returnFocus()
+    }
+  }, [root, children])
+
+  return React.createElement('div', {
+    ref: root,
+    children,
+    className: 'outline-none focus-trap',
+    tabIndex: -1,
+  })
```
lfades, line 63: Why use this over `<div ref={root} className="outline-none focus-trap" tabIndex={-1}>{children}</div>`?

tailwind.config.js
```diff
@@ -8,6 +8,11 @@ module.exports = {
       './pages/**/*.{js,ts,jsx,tsx}',
       './components/**/*.{js,ts,jsx,tsx}',
     ],
+    options: {
+      safelist: {
+        standard: ['outline-none'],
```
lfades, line 13: Why is it needed to stop PurgeCSS from removing this class?

lfades (changes requested):
Added some comments. Also please remove the `package-lock.json` that's currently included in the PR 🙏

lib/focus-trap.tsx
okbel: No reason, just taste for functions sometimes.

tailwind.config.js
okbel: 🤔  No need, we can remove it!

components/ui/Modal/Modal.tsx
okbel: I read somewhere that it actually there was not point in wrapping this into `useCallback` for performance reasons. Only, in case of multiple instantiations that resulted in the same functions. Will add it just in case.

okbel:
Done!

components/ui/Modal/Modal.tsx
lfades: Is not for performance in this case though, is to avoid using an old version of `onClose`, which is very hard to debug. We could have a linter to be more restrictive with that.

components/ui/Modal/Modal.tsx
```diff
@@ -7,45 +7,53 @@ import {
...
+        window.addEventListener('keydown', handleKey)
       } else {
         enableBodyScroll(ref.current)
       }
     }
     return () => {
+      window.removeEventListener('keydown', handleKey)
       clearAllBodyScrollLocks()
     }
   }, [open])
```
lfades, line 42:
```suggestion
  }, [open, handleKey])
```

`useCallback` was added to `handleKey` with the intention of adding it as a dep to this `useEffect`.

An alternative is you could move the `handleKey` function to be inside the `useEffect` and then add `onClose` as a dependency.

PR #149, opened 2021-01-07
Updates to improve lighthouse performance

dabit3:
This should bring Lighthouse to 100 on Best Practices. There are still some minor issues around accessibility but are more styling related I think.


okbel:
Lovely, thank you Nader!

okbel (approved):

PR #174, opened 2021-01-21
@hotfix/fixing brand id in use-search

mohammadou1:
to set searchParam brand, should check brandId and not categoryId


okbel (approved):

okbel:
Thank you, @mohammadou1 !

PR #177, opened 2021-01-23
Add cart item options like color and size

hlopes:
Hi,

I implemented a simple approach to display the item options in the Cart.
I'm not comfortable with the query parameter "?include=line_items.physical_items.options" being hardcoded on the get cart handler, any advice on this is welcome.

One thing that I noticed was that the helper getCurrentVariant didn't support just one defined option. For example, even if the user chose the color option without the size, still the first variant was going to be used.

Finally, no design was followed, just a simple output of the options.

Here is an example of the feature:

cart-with-options

Any feedback would be appreciated.

### DIFF
--- a/components/cart/CartItem/CartItem.tsx
+++ b/components/cart/CartItem/CartItem.tsx
@@ -8,6 +8,13 @@ import useUpdateItem from '@framework/cart/use-update-item'
 import useRemoveItem from '@framework/cart/use-remove-item'
 import s from './CartItem.module.css'
 
+type ItemOption = {
+  name: string,
+  nameId: number,
+  value: string,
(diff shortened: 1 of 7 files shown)


bbeldame:
You should add ?include=line_items.physical_items.options on remove, update and add too or it will have a buggy behavior when you update the cart in the sidebar.

hlopes:
Thanks for feedback. You're absolutely right.
I'll fix that.

okbel:
This looks good! I'm chatting with our design team to see if we can tweak it a little bit. Most of the variants are solved in the `agnostic` branch btw. Thank you 🙏

okbel (approved):

PR #178, opened 2021-01-24
OptionSelections type move in array

bbeldame:
I guess it's a typo in the code, I had this problem using optionSelections in a add cart on my project.


okbel:
This was addressed in the `agnostic` branch. Thank you!

PR #179, opened 2021-01-24
Issue/79 fallback image

leahtard:
Thanks for the feedback @okbel. I have updated the PR here :)


okbel:
Looks good! 

What do you think, @lfades? Only worried about CLS in case `src` gets populated later, but I don't think that could happen in any case 
 🤔

okbel (approved):

lfades (approved):
LGTM!

PR #180, opened 2021-01-24
add horizontal margin to pages when mobile screen

vczb:
When I access the pages with my cell phone it is difficult to read the text because there are no horizontal margins

about-page-without-margin

To solve this problem I added a small horizontal margin when accessing small screens

about-page-with-margin

I believe it got better

Thank you very much

Sincerely


tailwind.config.js
```diff
@@ -15,6 +15,13 @@ module.exports = {
     },
   },
   theme: {
+    screens: {
```
okbel, line 18: Remove this lines from config. These are the defaults. No need to add them

okbel: https://tailwindcss.com/docs/breakpoints

okbel (changes requested):

tailwind.config.js
vczb: I added these lines in the second commit because the verification link was breaking ...

I believe that the default did not work, but I can take it out to test it again.

Thanks

okbel (approved):

okbel:
Thanks!

PR #184, opened 2021-01-28
Fix the body scroll when the sidebar is open

hlopes:
Hi, I noticed that the body scroll wasn't locked when the sidebar was open, however has the logic for that.

I wasn't working because the Layout component is rerendered after any change in the UI Context, and after the sidebar is rendered the Modal component executes clearAllBodyScrollLocks() in his unmount function of useEffect invalidating the sidebar body scroll.

Another minor fix is, close the sidebar when the modal is open (for example pressing the account link inside the sidebar).


okbel (approved):

okbel:
Thanks!

PR #185, opened 2021-01-28
Remove duplicated css rules.

hlopes:
Remove duplicated css rules in base.css and fix invalid JSX props in SVGs.


okbel:
Thanks!

PR #194, opened 2021-02-05
Update sort filter texts on mobile

GuilleAngulo:
### Description
At **search** page, on mobile devices, the **sort** button shows the content of the query (_e.g._, **Sort: latest-desc**). I though that by changing SORT to a plain object it would be easier to access the value of the key (the label), without requiring iterating or filtering the entries. Not sure if is the best approach, hope it can be of some help :)

### Screenshots

|Before                    |After
| ---------------------------------- | ---------------------------------- |
| before| after |

### DIFF
--- a/components/search.tsx
+++ b/components/search.tsx
@@ -14,12 +14,12 @@ import useSearch from '@framework/product/use-search'
 import getSlug from '@lib/get-slug'
 import rangeMap from '@lib/range-map'
 
-const SORT = Object.entries({
+const SORT = {
   'trending-desc': 'Trending',
   'latest-desc': 'Latest arrivals',
   'price-asc': 'Price: Low to high',
   'price-desc': 'Price: High to low',
-})
+}
 
 import {
   filterQuery,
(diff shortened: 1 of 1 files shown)


okbel:
Beautiful! Thank you!

pages/search.tsx
```diff
@@ -377,7 +377,10 @@ export default function Search({
                   aria-haspopup="true"
                   aria-expanded="true"
                 >
-                  {sort ? `Sort: ${sort}` : 'Relevance'}
+                  {sort &&
+                  Object.prototype.hasOwnProperty.call(SORT, sort as string)
```
lfades, line 381: I think it's fine to do a simpler check here

lfades, line 381: I think it's fine to do a simpler check here

lfades (changes requested):
@GuilleAngulo Hi there and sorry for the late review!

Can you update the PR to the latest commit 🙏

GuilleAngulo:
hey @lfades 👋! I just made the changes and added a simpler check for the button label as you requested. Hopefully now everything is correct ... thanks =)

pages/search.tsx
GuilleAngulo: making the changes!

lfades (approved):
Thank you!

Issue #243, opened 2021-03-23
Sold Out button on PDP (possible solution)


DanielHirunrusme:
Currently, a user can add items to the cart when an item inventory is 0. Not good. Below is a possible solution. This does not solve variants available/sold out however. 

For Shopify provider, I added "totalInventory" to the query.

```javascript
const getProductQuery = /* GraphQL */ `
  query getProductBySlug($slug: String!) {
    productByHandle(handle: $slug) {
      id
      handle
      title
      productType
      vendor
      description
      descriptionHtml
      totalInventory
      options {
        id
        name
        values
      }
      priceRange {
        maxVariantPrice {
          amount
          currencyCode
        }
        minVariantPrice {
          amount
          currencyCode
        }
      }
      variants(first: 250) {
        pageInfo {
          hasNextPage
          hasPreviousPage
        }
        edges {
          node {
            id
            title
            sku
            selectedOptions {
              name
              value
            }
            priceV2 {
              amount
              currencyCode
            }
            compareAtPriceV2 {
              amount
              currencyCode
            }
          }
        }
      }
      images(first: 250) {
        pageInfo {
          hasNextPage
          hasPreviousPage
        }
        edges {
          node {
            originalSrc
            altText
            width
            height
          }
        }
      }
    }
  }
`
```

```javascript
export const productConnection = `
pageInfo {
  hasNextPage
  hasPreviousPage
}
edges {
  node {
    id
    title
    vendor
    handle
    description
    totalInventory
    priceRange {
      minVariantPrice {
        amount
        currencyCode
      }
    }
    images(first: 1) {
      pageInfo {
        hasNextPage
        hasPreviousPage
      }
      edges {
        node {
          originalSrc
          altText
          width
          height
        }
      }
    }
  }
}`
```

then in productView I added

```javascript
{product.totalInventory > 0 ? (
                <Button
                  aria-label="Add to Cart"
                  type="button"
                  className={s.button}
                  onClick={addToCart}
                  loading={loading}
                >
                  Add to Cart
                </Button>
              ) : (
                <Button
                  aria-label="Sold Out"
                  type="button"
                  className={s.button}
                  disabled
                >
                  Sold Out
                </Button>
              )}
```

Suppose this would need to be normalized for other providers. Also, adding option to continue selling if item is sold out.

okbel:
cc @cond0r

itsjzt:
[...]

PR #246, opened 2021-03-27
Shopify Provider Fixes

cond0r:
- Fixed the product query to fetch only available products. Product must have at least one variant in stock, if track quantity is active. Otherwise, product doesn't appear #243 
- Add product descriptionHtml  with fallback to plain text description on ProductView (Since the description supports HTML on Shopify Product Editor) #258 
- User can bulk enable to track or not to track quantity, or to sell or not items when out of stock directly from Shopify. From commerce provider perspective this is covered by checking availableQuantity from product variant to set isInStock, and check availableForSale.

Screenshot 2021-03-27 at 20 30 03

If variant has inventory tracking and availability 0 we display out of stock and disable button

Screenshot 2021-03-27 at 18 58 02

If user clicks on "Continue selling when out of stock" add to cart button is active again

 Screenshot 2021-03-27 at 18 58 11

Screenshot 2021-03-27 at 18 58 41


davemooreuws:
@okbel Can you please comment on the reason for this PR being closed?

PR #251, opened 2021-03-29
Fix: type "any" for correct type 

niko20:
Hi just a little type fix in _app.tsx

before: 
```
const Layout = (Component as any).Layout || Noop
```

After: 
```
type LayoutT = FC & { Layout: typeof Layout }
 const Layout = (Component as LayoutT).Layout || Noop
```

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -4,13 +4,15 @@ import 'keen-slider/keen-slider.min.css'
 
 import { FC, useEffect } from 'react'
 import type { AppProps } from 'next/app'
-import { Head } from '@components/common'
+import { Head, Layout } from '@components/common'
 import { ManagedUIContext } from '@components/ui/context'
 
-const Noop: FC = ({ children }) => <>{children}</>
+type LayoutT = FC & { Layout: typeof Layout }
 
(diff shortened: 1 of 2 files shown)


pages/_app.tsx
```diff
@@ -4,13 +4,15 @@ import 'keen-slider/keen-slider.min.css'
...
 import { ManagedUIContext } from '@components/ui/context'
 
-const Noop: FC = ({ children }) => <>{children}</>
+type LayoutT = FC & { Layout: typeof Layout }
 
+const Noop: FC = ({ children }) => <>{children}</>
+cl
 export default function MyApp({ Component, pageProps }: AppProps) {
-  const Layout = (Component as any).Layout || Noop
+  const Layout = (Component as LayoutT).Layout || Noop
```
lfades, line 15: In this case the type `any` is fine, we already know that `Component` is a react component, but the code isn't interested in the value of `Layout` or whether it exists or not (as it has a fallback), and we also don't want to enforce it to be the `Layout` component.

lfades:
@niko20 Hi there and thank you for the PR. Closing per https://github.com/vercel/commerce/pull/251#discussion_r679634165

Issue #224, opened 2021-03-11
Unify the Node.js API for providers


lfades:
Currently we have API methods that work under `/api/bigcommerce/*` that applies for all providers. The API used for the Node.js side of the commerce demo should be unified in a similar way of the agnostic provider we currently have for the UI hooks.

Vandivier:
@lfades I support this but let's also think one step higher in abstraction: There are some API that go beyond providers and would be used for any back end. For example, `isAllowedMethod`.

ref: https://github.com/vercel/next.js/discussions/25599

As a side note, I find the higher-order hook implementation like `swell/.../use-signup.tsx` to be really hard to reason about. Just a note that 'unify the API' doesn't need to be done with a compositional approach. isAllowedMethod, for example, could simply be pasted in `framework/utils` and the other files could directly import it without composition.

lfades:
@Vandivier `isAllowedMethod` was moved to be inside `commerce` so that's already the case, on the note about compositional approach that's a requirement because providers don't match each other's features, each provider adds what they can do

PR #252, opened 2021-03-30
[WIP] Node.js provider for the API

lfades:
Fixes https://github.com/vercel/commerce/issues/224

This PR refactors the way the Node.js API of commerce and the UI interact with each other. In general it does the following:

- API endpoints now share the same path, independent of the provider, and the endpoint now lives in the core, with each provider now adding handlers for that endpoint.
- API operations now live under a common `commerce` object, instead of being independently imported, they're still independent, but this way it feels more like a Node.js API
- Commerce core types now properly include the types for everything, previously we had different types for API endpoints, API operations and UI hooks, now they all come from the same place, this also makes our types less complicated and tricky to use

Next steps: testing and updating the other providers to follow this new structure, BigCommerce is 100% right now.


Lmillan123 (commented):
- [ ]

uchitwai (commented):
Ok

PR #255, opened 2021-03-31
Fix deployment for certain product options in BigCommerce

ikoruk:
add default values to fix case where there are no values for product option (like text input)


ikoruk:
This is a very simple fix, not sure why it was not authorized...

ikoruk:
I am still having the same issue when trying to deploy now, so it would be good to accept this

lfades (approved):
@ikoruk Thank you!

Issue #258, opened 2021-04-04
[ENHANCEMENT] Shopify Provider // descriptionHtml for product description (solution included)


DanielHirunrusme:
Realized that the product description pulled from Shopify is stripped of HTML, therefore loses formatting, links, etc. Is this intentional?

imo, think it makes more sense from an editor perspective to keep html formatting. The developer can use serializers to reformat if needed. Edit below:

Added descriptionHtml to normalize.ts

```javascript
export function normalizeProduct(productNode: ShopifyProduct): Product {
  const {
    id,
    title: name,
    vendor,
    images,
    variants,
    description,
    descriptionHtml,
    handle,
    priceRange,
    options,
    ...rest
  } = productNode

  const product = {
    id,
    name,
    vendor,
    description,
    descriptionHtml,
    path: `/${handle}`,
    slug: handle?.replace(/^\/+|\/+$/g, ''),
    price: money(priceRange?.minVariantPrice),
    images: normalizeProductImages(images),
    variants: variants ? normalizeProductVariants(variants) : [],
    options: options
      ? options
          .filter((o) => o.name !== 'Title') // By default Shopify adds a 'Title' name when there's only one option. We don't need it. https://community.shopify.com/c/Shopify-APIs-SDKs/Adding-new-product-variant-is-automatically-adding-quot-Default/td-p/358095
          .map((o) => normalizeProductOption(o))
      : [],
    ...rest,
  }

  return product
}
```

Added descriptionHtml to get-all-products-query.ts
```javascript
export const productConnection = `
pageInfo {
  hasNextPage
  hasPreviousPage
}
edges {
  node {
    id
    title
    vendor
    handle
    description
    descriptionHtml
    totalInventory
    priceRange {
      minVariantPrice {
        amount
        currencyCode
      }
    }
    images(first: 1) {
      pageInfo {
        hasNextPage
        hasPreviousPage
      }
      edges {
        node {
          originalSrc
          altText
          width
          height
        }
      }
    }
  }
}`

export const productsFragment = `
products(
  first: $first
  sortKey: $sortKey
  reverse: $reverse
  query: $query
) {
  ${productConnection}
}
`

const getAllProductsQuery = /* GraphQL */ `
  query getAllProducts(
    $first: Int = 250
    $query: String = ""
    $sortKey: ProductSortKeys = RELEVANCE
    $reverse: Boolean = false
  ) {
    ${productsFragment}
  }
`
export default getAllProductsQuery
```

[...]

PR #274, opened 2021-04-14
Shopify product description html

cond0r:
Add descriptionHtml with fallback to plain text description in ProductView #258 
https://commerce-git-productdescriptionhtml-cond0r.vercel.app/product/vans-authentic-lo-pro-burgandy-white


okbel (approved):
Looks good!

PR #306, opened 2021-05-06
Bigcommerce options fix

lewis52ravens:
CartItem.tsx attempts to access the `options` field of `LineItem`, however this field is only available with the Shopify provider.
This fix:
- adds the `options` field to the core `LineItem` type, so it can be accessed regardless of provider.
- removes the `LineItem` extension from the Shopify types file, since it is no longer needed
- updates the `normalizeLineItem` function for BigCommerce to copy `item.options` to the `options` field of the returned `LineItem`
- Sorts the `options` array before displaying them in CartItem.tsx, so that options will always appear to the user in the same order (i.e. [size], [color])

### DIFF
--- a/components/cart/CartItem/CartItem.tsx
+++ b/components/cart/CartItem/CartItem.tsx
@@ -116,15 +116,34 @@ const CartItem = ({
         </Link>
         {options && options.length > 0 ? (
           <div className="">
-            {options.map((option: ItemOption, i: number) => (
-              <span
-                key={`${item.id}-${option.name}`}
-                className="text-sm font-semibold text-accents-7"
-              >
-                {option.value}
-                {i === options.length - 1 ? '' : ', '}
-              </span>
-            ))}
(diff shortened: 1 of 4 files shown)


components/cart/CartItem/CartItem.tsx
```diff
@@ -116,15 +116,34 @@ const CartItem = ({
...
+            {/** attempt to sort the options array alphabetically by option name, that way options will always appear in the same order */}
+            {options
+              .sort((first: any, second: any) => {
+                if (first.name && second.name) {
+                  return first.name < second.name
+                    ? -1
+                    : first.name > second.name
+                    ? 1
+                    : 0
+                } else {
```
okbel, line 128: This logic should be done at the provider level, not in the view. Could you move it?

lfades:
Closing due to conflicts and Bel's feedback not being solved yet. If you're still interested on working in this PR let me know and I'll open it!

Issue #291, opened 2021-04-26
How to approach pricing at the deepest option level?


loan-laux:
While building a provider for Reaction Commerce, I stumbled onto an issue that I think many other developers will have with other APIs.

The structure for product variants is as follows:

```
iPhone 12 (main product)
  |
  |
  |--- Black (variant)
  |       |
  |       |--- 64GB (sub-variant) - $699
  |       |--- 128GB (sub-variant) - $799
  |       |--- 256GB (sub-variant) - $899
  |
  |
  |--- Purple (variant)
  |       |
  |       |--- 64GB (sub-variant) - $799
  |       |--- 128GB (sub-variant) - $899
  |       |--- 256GB (sub-variant) - $999

```

With this structure, the sub-variants (or the sub-sub-variants, or whichever deepest level) is the one that bears a price. There is no price at the variant level (unless it's the deepest), nor at the main product level.

As I understand it, the approach of Next Commerce is to have the price at the variant level (which is expected to ideally be colors), and then however many `MultipleChoiceOption`s we want, but which won't influence the price of the item.

Is that correct? If so, how would you go about making a variant structure like Reaction's work with Next Commerce without having to touch too many core files?

loan-laux:
Turns out I was just misunderstanding how this whole `MultipleChoiceOption` worked. Maybe it should be better documented, but at least we got it to work on our provider, so I'm closing this.

PR #313, opened 2021-05-18
Allow a product variant to have a different price

royderks:
This will allow a product variant to have a different price than the product itself

### DIFF
--- a/components/product/ProductView/ProductView.tsx
+++ b/components/product/ProductView/ProductView.tsx
@@ -19,14 +19,17 @@ interface Props {
 
 const ProductView: FC<Props> = ({ product }) => {
   const addItem = useAddItem()
-  const { price } = usePrice({
-    amount: product.price.value,
-    baseAmount: product.price.retailPrice,
-    currencyCode: product.price.currencyCode!,
-  })
   const { openSidebar } = useUI()
   const [loading, setLoading] = useState(false)
   const [choices, setChoices] = useState<SelectedOptions>({})
+
+  const variant = getVariant(product, choices)
+
+  const { price } = usePrice({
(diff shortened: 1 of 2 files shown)


okbel (approved):

loan-laux:
Looks like this would solve https://github.com/vercel/commerce/issues/291.

okbel:
Does this only add price to a variant? Could you delete the other changes? Looks like it's failling. So, maybe we need to also update the UI as well.

royderks:
> Does this only add price to a variant? Could you delete the other changes? Looks like it's failling. So, maybe we need to also update the UI as well.

There are no other changes. This adds a new field to the type for product variants with a price, if the provider supplies that field it will render a price

components/product/ProductView/ProductView.tsx
```diff
@@ -19,15 +19,18 @@ interface Props {
...
   const { openSidebar } = useUI()
   const [loading, setLoading] = useState(false)
   const [choices, setChoices] = useState<SelectedOptions>({})
 
+  const variant = getVariant(product, choices)
+
+  const { price } = usePrice({
+    amount: variant?.price!.value || product.price.value,
+    baseAmount: variant?.price!.retailPrice || product.price.retailPrice,
+    currencyCode: variant?.price!.currencyCode! || product.price.currencyCode!,
```
cond0r, line 31: @royderks this fails at bigcommerce because expects variant price to not be null, so replace `variant?.price!.` with  `variant?.price?. `

lfades:
Closing for now as Bigcommerce is failing, there are conflicts, and the PR hasn't been updated in months. @royderks feel free to update it and I'll re-open the PR. Thank you.

PR #333, opened 2021-05-27
fix update cart item

greph:
* updates the correct cart item for increment/decrement quantity


okbel:
Perfect! For that delay to update the UI we will improve it on our end 👍

okbel (approved):

PR #338, opened 2021-05-28
Fix vendure variants

michaelbromley:
Fixes selection of the correct product variant for the Vendure provider


okbel:
Works perfectly 🙌

PR #339, opened 2021-05-28
Improved Categories

okbel:
- Adding initial types for Categories.
- Navbar now shows a slice of the categories instead of being hardcoded.


framework/bigcommerce/common/get-site-info.ts
```diff
@@ -94,11 +96,20 @@ async function getSiteInfo({
...
-  const categories = data.site?.categoryTree
+
+  let categories = data!.site!.categoryTree?.map(
+    ({ entityId, name, path }: any) => ({
+      id: `${entityId}`,
+      name,
+      slug: getSlug(path),
+      path,
+    })
+  )
```
lfades, line 107: Let's make this a normalize function, like the one we have for cart data, and add a comment with a TODO on top of the `any` type, we have BC types that should be able to match here.

lfades (approved):

PR #343, opened 2021-05-30
Move checkout forms to provider

loan-laux:
This PR moves the checkout form components to the providers, so that each provider can provide its own logic and UI for payment and shipping. This way, providers can decide to use Stripe Elements or whatever works best with their backend.


tsconfig.json
```diff
@@ -19,6 +19,7 @@
       "@utils/*": ["utils/*"],
       "@config/*": ["config/*"],
       "@assets/*": ["assets/*"],
+      "@components/checkout/*": ["framework/shopify/components/checkout/*"],
```
lfades, line 22: I don't think this is required, the way we currently import hooks should also work for components.

lfades (changes requested):
Hi @loan-laux👋 

Before moving on, we'll me merging the nodejs agnostic PR tomorrow most likely: https://github.com/vercel/commerce/pull/252 - This would change the way we use the provider and integrate new stuff to it, like the checkout.

We also need to spec this first, I think the checkout should also work like a hook, meaning a `useCheckout` hook that returns checkout info and that may also handle form submission, validation and other things unrelated to design, similar to the way https://react-hook-form.com/ works. Once that's done we can define a common Checkout component if needed, but we shouldn't be starting there.



Let's spec this properly

loan-laux:
Thanks for taking a look, @lfades! Sure, I'll keep an eye on #252 today. I also agree on a `useCheckout` hook. Maybe even also a `useAvailablePaymentMethods` hook since a lot of e-commerce APIs will have a way to query all the available payment methods. And then render each payment method from the provider itself.

As you said, we need to spec this first. Let's wait for the Node.js provider API to be merged and then I'll come back with a better-defined idea.

Issue #345, opened 2021-05-31
remove extraneous code related to ProductCard `variant="slim"`


Vandivier:
1. `variant="slim"` provides a default height and width but we always override it w imgProps

So can we just remove the imgProps in those cases?
related discussion: https://github.com/vercel/commerce/discussions/281

PR #348, opened 2021-05-31
feat: concise card code

Vandivier:
closes https://github.com/vercel/commerce/issues/345

imgProps default to 320 within the component, so redeclaration in index.tsx isn't needed.


Vandivier:
Hi @okbel - I think you are familiar with that ProductCard default height/width code.

lfades (approved):
Thank you @Vandivier !

PR #351, opened 2021-06-01
Added Vendure search input variables for `getAllProducts`

tomsvogel:


okbel:
@michaelbromley I let Michael review this one

michaelbromley:
1. It needs updating with the current master after that big refactor. Toms - pull the latest version and you'll find the get-all-products.ts file now in the `/vendure/api/operations/` dir.
2. It looks fine technically - as in it will pass the correct data to the Vendure API. The only thing I'm not sure of is whether the Vercel team intend to be able to define arbitrary variables in the operations API like this, or are you planning a common interface so that the front-end components won't need to change when switching providers? Because as it is, this PR includes exposes some quite Vendure-specific variables.

lfades:
Closing the PR as it's outdated and doesn't fit our current structure.

PR #355, opened 2021-06-04
Improvements

shuding:
- fix public asset URLs
- fix type import
- use esm version of npm dependency
- parallelize SSG data fetching


zaiste:
@shuding an interesting change. Have you had a chance to measure how much the data parallelisation helps here? 

I wonder if it wouldn't be more straightforward / explicit to group these _extracted_ promises within `await Promise.all(...)`

shuding:
@zaiste I don't have accurate numbers (and it varies for different CMS providers), but it speeds up each `getStaticProps` by more than 1.5s for me on my laptop.

okbel:
These are already in :galatic-relief:

PR #277, opened 2021-04-14
Saleor provider for Next.js Commerce

patrys:
⚠️  This is not a working implementation. I'm opening a draft PR to give us a place to review and discuss the necessary changes.

**Status**

* [x] CommerceProvider
* [x] Schema & TS types
* [x] API Operations - Get all collections
* [x] API Operations - Get all pages
* [x] API Operations - Get all products
* [x] API Operations - Get page
* [x] API Operations - Get product
* [ ] API Operations - Get Shop Info
* [x] Hook - Add Item
* [x] Hook - Remove Item
* [x] Hook - Update Item
* [ ] Hook - Get Cart (missing tax & shipping)
* [ ] Auth and customer information
* [ ] Product attributes - Size, Colours
* [ ] Custom checkout
* [ ] Typing
* [ ] Tests

**Instructions**

To get it to run you will need to copy `framework/saleor/.env.template` to `.env.local` and fill in the URL and the channel slug.

You will also need to disable all features in the global `commerce.config.json` (Next.js Commerce currently ignores the feature flags declared by individual providers).

PR #356, opened 2021-06-05
Updated Saleor Provider

zaiste:
This PR supersedes the #277 It also incorporates the changes introduced by #252 Overall, this PR provides a minimal working example showing how to integrate Saleor with the Next.js Commerce platform. 

cc @patrys @okbel 

Additionally, I've updated a couple of dependencies (`package.json`) - leaving Next.js at `10.0.9-canary.5` because of issues that prevent images to be properly displayed cc @lfades


lfades:
@zaiste What's the issue that prevents us from being on the latest Next.js version? 10.0.9 is quite behind

zaiste:
@lfades Images didn't appear in `production` for any version of Next.js after the current one if I recall correctly. I can investigate further as I'd also prefer to be on the latest Next.js version here.

lfades:
@zaiste Is that the case for the other providers? 🤔

zaiste:
@lfades I haven't yet checked other providers. For clarity, currently, in this repository on the `master` branch, you use this older Next.js version i.e. `10.0.9-canary.5` ([ref](https://github.com/vercel/commerce/blob/master/package.json#L37)). In this PR, I decided to leave the Next.js version __unchanged__. We could address the Next.js upgrade in the following PR to make sure it works correctly for all providers across the board.

PR #364, opened 2021-06-15
optionsSelections takes an array of options objs

ryanford:
When adding an item to cart with product options, the `optionSelections` member should accept an array of product option objects. Currently this causes an error if you try to add products with options that match commerce's `CartItemBody` type
<https://developer.bigcommerce.com/api-reference/storefront/carts/cart-items/addcartlineitem>


lfades (approved):
LGTM. Thank you @ryanford !

PR #390, opened 2021-06-23
Remove tsconfig.js and add back yarn.lock

lfades:
- Files inside `framework` shouldn't be importing files from outside, so I'm undoing that change that added `tsconfig.js` in https://github.com/vercel/commerce/pull/385
- A yarn.lock or package lock should be existing in the project.


okbel (approved):

sokra:
The tsconfig.js file is needed for next-live

PR #391, opened 2021-06-25
Vendure - Fetcher - Globally throw errors

gkielwasser:
@michaelbromley 

Globally throw errors from the Vendure fetcher when we detect errors in the response body.

It allows not to have to check it in each hooks. We can also safely destructure the response from the hook.

Currently, in case of an unhandled exception in the server, something like what we do in the `use-cart.tsx` hook would break. `activeOrder` property wouldn't exist on data response.
```
export const handler: SWRHook<GetCartHook> = {
  ...
  async fetcher({ input: { cartId }, options, fetch }) {
    const { activeOrder } = await fetch<ActiveOrderQuery>(options)
    return activeOrder ? normalizeCart(activeOrder) : null
  },
  ...
}

### CODE REVIEW

**Comment by michaelbromley** [comment]
@okbel the author and I discussed this in the Vendure slack - this improves fault tolerance for the Vendure API.

File: framework/vendure/fetcher.ts [hunk 1]
```diff
@@ -9,12 +9,18 @@ async function getText(res: Response) {
   }
 }
 
+async function fetcherError(options: {
+  status: number
+} & ErrorProps) {
+  return new FetcherError(options)
+}
+
```
lfades, line 17: `FetcherError` already has the `status` prop.

lfades (changes requested):
Thank you for the PR! Added one comment

framework/vendure/fetcher.ts
gkielwasser: Hi,
Not sure it's what you meant but I just removed unnessary new function

lfades (approved):
Thank you!

PR #412, opened 2021-07-09
Per Provider Configuration: Ability to customise SEO, et al.

zaiste:
## Per-Provider Configuration
Right now, the config is only about SEO. By moving it to  the `@framework` directory, it can be customized on a per-provider basis. This way each provider website can provide a slight variation in wording, og:image etc.

## Apple M1 compatibility
It's a workaround to make Next.js work on Apple M1 with Node.js 16. It does not optimise the images in development. This `<Image>` alias should be removed once the https://github.com/vercel/next.js/issues/24421 is resolved.


goncy:
Hey @zaiste thanks for the PR, https://github.com/vercel/next.js/issues/24421 is resolved now so it should be safe to remove the Image component hack, regarding the config for SEO per provider, a file should be added to every provider so that import doesn't break in other deploys. @lfades what do you think about importing it directly from framework?

goncy:
Great @zaiste, I will close this for now but we will check this closer on our side, thanks for the help 🙌

PR #416, opened 2021-07-16
Rename framework to providers.

ItzLue:


lfades:
@ItzLue Hi there and thank you for the PR. Currently we don't want to rename the folders from framework to providers, feel free to create an issue if have a good reason to make the change

PR #417, opened 2021-07-21
Add note in Vendure readme on correct server config for sessions

michaelbromley:
Avoid confusion by making sure devs know how to set up sessions to work properly with this provider.


lfades (approved):
Thank you!

Issue #431, opened 2021-08-03
Shopify Checkout Throttling and Queue


ottomated:
This repository is missing two critical features required for large Shopify stores:

> **Checkout-level throttle**
Shopify limits the amount of checkouts that can be created on the Storefront API per minute. If an API client exceeds this throttle, then a 200 Throttled error response is returned. Shopify recommends designing your app to be resilient to this scenario. For example, you could implement a request queue with an exponential backoff algorithm.

> **/throttle/queue**
Shopify redirects users in the checkout flow to `/throttle/queue` if the store is under high traffic.

Without these features, two things will happen. The API silently fails when creating a checkout session, meaning that addToCart requests also silently fail. Also, during the checkout flow, the user will be redirected to a nonexistent URL.

DanielHirunrusme:
This is an interesting subject. If a store is facing this issue during sales then bots may likely be involved. 

Not only does Shopify Plus have a throttle but also a checkpoint option to solve a captcha before receiving a checkout url.

Obviously, some bots can still get around this checkpoint. It’s an uphill battle and Shopify Plus is constantly adding patches to bot squash and handle traffic.

Question is: During these types of sales do you want to be responsible for scaling the site to handle this amount of traffic/bots or allowing Shopify by using a theme?

balakedev:
> If a store is facing this issue during sales then bots may likely be involved. 

Lot's of shopify plus stores hit peaks of more than 4000 concurrent users, particularly merchants that have product releases or drops. I would say this is something the vercel commerce team needs to address pretty quickly. There are work arounds that you can do in the existing project to resolve this issue that don't involve a queue. But it would be better if the core did actually use cart instead of checkout.

muhammad-saleh:
Referencing this here:
https://github.com/vercel/commerce/issues/430

ottomated:
> This is an interesting subject. If a store is facing this issue during sales then bots may likely be involved.

I actually created this issue because I manage https://ludwig.gg, and ran into these issues yesterday during a merch release. Especially since by default, the repository creates a cart session upon the first `useCart` call, it doesn't take too much traffic to trigger throttling. Perhaps if there isn't an existing cart, the framework should wait until the first add, remove, or update request to actually create the cart session.

> Question is: During these types of sales do you want to be responsible for scaling the site to handle this amount of traffic/bots or allowing Shopify by using a theme?

[...]

PR #432, opened 2021-08-05
Shopify: create checkout on add to cart

cond0r:
Creates a checkout only when user first time interacts with the cart, not on the first visit of the store #431

### DIFF
--- a/framework/bigcommerce/cart/use-add-item.tsx
+++ b/framework/bigcommerce/cart/use-add-item.tsx
@@ -29,16 +29,18 @@ export const handler: MutationHook<AddItemHook> = {
 
     return data
   },
-  useHook: ({ fetch }) => () => {
-    const { mutate } = useCart()
+  useHook:
+    ({ fetch }) =>
+    () => {
+      const { mutate } = useCart()
 
-    return useCallback(
-      async function addItem(input) {
-        const data = await fetch({ input })
-        await mutate(data, false)
-        return data
-      },
-      [fetch, mutate]
-    )
-  },
+      return useCallback(
+        async function addItem(input) {
+          const data = await fetch({ input })
+          await mutate(data, false)
+          return data
+        },
+        [fetch, mutate]
+      )
+    },
 }
--- a/framework/bigcommerce/cart/use-add-item.tsx
+++ b/framework/bigcommerce/cart/use-add-item.tsx
@@ -29,18 +29,16 @@ export const handler: MutationHook<AddItemHook> = {
 
     return data
   },
-  useHook:
-    ({ fetch }) =>
-    () => {
-      const { mutate } = useCart()
+  useHook: ({ fetch }) => () => {
+    const { mutate } = useCart()
 
-      return useCallback(
-        async function addItem(input) {
-          const data = await fetch({ input })
-          await mutate(data, false)
-          return data
-        },
-        [fetch, mutate]
-      )
-    },
+    return useCallback(
+      async function addItem(input) {
+        const data = await fetch({ input })
+        await mutate(data, false)
+        return data
+      },
+      [fetch, mutate]
+    )
+  },
 }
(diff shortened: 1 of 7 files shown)


ottomated:
This is promising but I would say it needs both an option to override this behavior and for this to happen on any cart-related action.

cond0r:
> This is promising but I would say it needs both an option to override this behavior and for this to happen on any cart-related action.

The other cart related  are not active since the cart is empty / null. So you cannot click delete or update. And i just checked. I deleted the cart cookie, and you click on cart again, and the cart was empty, and i was able to click on add to cart, and the cookie & checkout was created again.

framework/bigcommerce/cart/use-add-item.tsx
```diff
@@ -29,16 +29,18 @@ export const handler: MutationHook<AddItemHook> = {
 
     return data
   },
-  useHook: ({ fetch }) => () => {
-    const { mutate } = useCart()
+  useHook:
```
goncy, line 32: Can you revert the changes for this file?

tsconfig.json
```diff
@@ -23,8 +23,8 @@
       "@components/*": ["components/*"],
       "@commerce": ["framework/commerce"],
       "@commerce/*": ["framework/commerce/*"],
-      "@framework": ["framework/local"],
-      "@framework/*": ["framework/local/*"]
+      "@framework": ["framework/shopify"],
```
goncy, line 26: Can you revert this changes?

goncy (changes requested):

tsconfig.json
cond0r: @goncy done

lfades (changes requested):
Hi there and thank you for the PR 👋 

Changes look good to me but the checkout isn't working in the demo: https://commerce-shopify-l41k21i4l-vercel-solutions-vtest314.vercel.app/

cond0r:
> Hi there and thank you for the PR 👋
> 
> Changes look good to me but the checkout isn't working in the demo: https://commerce-shopify-l41k21i4l-vercel-solutions-vtest314.vercel.app/

I think it as something to do with preview URL's you can see it did the same for the Implement custom checkout one https://commerce-shopify-git-implement-bfb334-vercel-solutions-vtest314.vercel.app/  but is ok in production one

cond0r:
@lfades now should work. On #487 the handler for checkout was changed to be on submit-checkout file which happens on POST, but since the /checkout is a redirect to /api/checkout, it has GET headers, so changed it to be on get-checkout handler. But what is strange, is that also BigCommerce should require the same changes to work, but works in the production URL now (same with Shopify). But it doesn't work on the preview URL's from the PR for example BigCommerce https://commerce-bigcommerce-git-imple-57f06f-vercel-solutions-vtest314.vercel.app/checkout

goncy:
> @lfades now should work. On #487 the handler for checkout was changed to be on submit-checkout file which happens on POST, but since the /checkout is a redirect to /api/checkout, it has GET headers, so changed it to be on get-checkout handler. But what is strange, is that also BigCommerce should require the same changes to work, but works in the production URL now (same with Shopify). But it doesn't work on the preview URL's from the PR for example BigCommerce https://commerce-bigcommerce-git-imple-57f06f-vercel-solutions-vtest314.vercel.app/checkout

You are right @cond0r, will take a look at this and get back at you so we can merge this 🙌

goncy:
@cond0r a PR has been merged to solve this, you can get back to use get-checkout as submit-checkout is not required anymore, thanks! Let me know once its done and I can merge this 🙌

cond0r:
> @cond0r a PR has been merged to solve this, you can get back to use get-checkout as submit-checkout is not required anymore, thanks! Let me know once its done and I can merge this 🙌

@goncy done, you can check again

goncy (approved):

goncy:
Merged, thanks @cond0r!

PR #438, opened 2021-08-10
Shopify Cart API & Wishlist

cond0r:
- use the new Storefront Cart API instead of Checkout (available only under unstable api version at the moment)
- implement Wishlist  (Since Cart API has no limitations, managed to add Wishlist functionality by creating another parallel cart used just to store items for the Wishlist and storing that id into a cookie same as cart). Since there is no need for customer when creating Wishlist this way, i allowed the add to Wishlist without requiring authentification if wishlistForGuests is set to true in commerce config.

Example: https://commerce-klkfzby0b-cond0r.vercel.app


cond0r:
> One thing worth noting on using the new Shopify Cart API for Wishlist is that they wouldn't necessarily persist indefinitely:
> 
> > Unused and abandoned carts are auto-expired by Shopify.
> 
> Source: https://shopify.dev/custom-storefronts/cart#considerations

Most of times i ever used a Wishlist, was to browse a catalog, and when i see something i like (directly from the grid list) to click that heart icon quickly and add into a temporary cart that i can use later to filter and decide what to add to cart at the end of my site session.  Otherwise, i have to open a new tab with items i like, or click & go to each product i like, click add to cart, and then go back to a list that maybe i have to scroll down to go where i was before, or click "load more" again when i have a large catalog. I think this UX is better (even if that cart expires in two weeks or a month) than consider as a Wishlist, something that i have to login to use it & have indefinitely (where in Shopify there is no such thing at the moment).

thomasbritton:
It's an interesting take on a 'wishlist', but I feel like this is a bit of a anti-pattern as I feel this project is an offering of what the Provider offers out of the box and as Shopify does not offer Wishlist functionality out of the box that this shouldn't be included as it gives the impression to people not used to Shopify in thinking that Wishlist functionality is standard when it isn't and will likely come into unexpected issues like the example that @juanferreras provided.

Also I think end-user expectations of a Wishlist would offer more features such as public/private lists, social sharing options, multiple wishlists etc.

PR #467, opened 2021-09-01
Run prettier autofix on all files.

danielgent:
Hey all :wave: 

Prettier is really helpfully set up on this repo *but* as it has not been run on all the files, that means our PRs are very messy as every time we touch a new file we get prettier formatting fixes in the diff :smile: 

I could just do this on my own private fork *but* that means I won't be able to keep up-to-date with this upstream.


goncy:
Hey @danielgent, thanks for the PR! This got outdated already, I will ensure to run it again and push it, thanks 🙌

PR #475, opened 2021-09-08
Update normalize.ts

pfcodes:
add missing options property to `normalizeLineItem`


lfades (changes requested):
@pfcodes Hi there and thank you for the PR 👋 

Is `item.options` being returned by Bigcommerce's item? its type is `any` so it could not be there and this change might have no effect at all.

PR #482, opened 2021-09-12
Fix next/link should be used with anchor as children #330

### LINKED ISSUE #330: next/link should be used with anchor as children
There are some places where `<Link>` is used without `<a>` as the children, this is bad for accessibility because links should be links. An example of this is the cart item, where the image and text go to the product page but without using an anchor, and right now the sidebar closes as soon as you click, instead of waiting for the navigation.

vczb:
Hi, I'm sending this PR to solve the issue "https://github.com/vercel/commerce/issues/330".

However, as I mentioned in the issue #330 , the problem of clicking and closing instantly is not due to the lack of the <a> tag, but because of the `closeSidebarIfPresent` method added to the click.

In this PR I'm just fixing the accessibility issue by adding the missing tag

Thank you and I'm available

close #330


lfades (approved):
I love it. Thank you!

PR #485, opened 2021-09-14
Update product.ts

pfcodes:
Pulls option entityId. 

This allows for adding a cart item with `option_selections` without having to rely on the complex "getProductVariant" helper function.

Screen Shot 2021-09-13 at 8 58 38 PM


lfades (changes requested):
@pfcodes Hi there 👋 

Is this PR complete and ready for review? Looks like it's only a change to a query, but the field isn't being used

PR #487, opened 2021-09-15
Implement custom checkout

goncy:
This PR adds basic functionality to cover custom checkout. The work on this PR can be divided as:

* Custom checkout core implementation: Mostly work done on `/framework/commerce`
* UI changes: Done in the `components` folder
* Provider related changes: Mostly added files on `checkout` folders to be correctly imported by the core

To try this out, ensure to have `"customCheckout": true` set in your `commerce.config.json`, this will enable a "Proceed to checkout" button in the cart overview sidebar instead of a redirection.

No provider is using this functionality yet but `ordercloud` provider is no the works so this PR blocks its development. You can checkout to `ordercloud-provider` branch to  see this implementation working.

Key points:
Related to UI, `CheckoutSidebarView.tsx` now uses a new hook called `useCheckout` that gets the checkout information from an API route, that response will contain `hasPayment` and `hasShipping` and a check will show instead of the chevron for true properties. The "Proceed to checkout" button will only be enabled when both properties are `true`. Validations should happen on the provider but modifying components and handle validations is not yet implemented in this iteration.

Here you can see how the checkout is being shown in different states:

image
> No steps completed

image
> Payment method set

image
> Payment and shipping set

Data schemas are not fully decided yet and most of the code is boilerplate oriented.


pfcodes:
Thanks for this. But `useCheckout` and `useAddCard` have no logic in their implementations.. at least for BigCommerce.

lfades (approved):
LGTM!

PR #518, opened 2021-10-04
Fix: Product Slider Firefox Layout

MarkWissinkTekton:
- Remove height style from product slider component
- Resolves an issue where the product thumbnails aren't shown on Firefox since the height attribute was causing the view to overflow


goncy:
Hey @MarkWissinkTekton thanks for this! It has been a while and it looks solved now
image

Feel free to reopen if its still an issue for you, thanks!

Issue #479, opened 2021-09-12
[BC] Feature request - Integration of the route /subscriptions


Florian-crg:
https://developer.bigcommerce.com/api-reference/storefront/storefront-subscriptions/subscription/postsubscriptions

kogan007:
Super easy to implement on your own. In framework/bigcommerce create api endpoint subscribe and make it somewhat similar to this 

`const { data } = await config.storeApiFetch('/v3/customers/subscribers', {
      method: 'POST',
      body: JSON.stringify(
        {
          ...(firstName && { first_name: firstName }),
          ...(lastName && { last_name: lastName }),
          email,
        },
      ),
    })`

I made name and last name conditional because BC API doesn't require them

Then just make a use-subscribe hook 
`import { useCallback } from 'react'
import type { MutationHook } from '@commerce/utils/types'
import { CommerceError } from '@commerce/utils/errors'
import useSubscribe, { UseSubscribe } from '@commerce/customer/use-subscribe'
import type { SubscriberHook } from '../types/subscribe'


export default useSubscribe as UseSubscribe<typeof handler>

export const handler: MutationHook<SubscriberHook> = {
  fetchOptions: {
    url: '/api/subscribe',
    method: 'POST',
  },
  async fetcher({ input: { email, firstName, lastName }, options, fetch }) {
    if (!(email)) {
      throw new CommerceError({
        message:
          'An email is required to subscribe',
      })
    }

    return fetch({
      ...options,
      body: { email, firstName, lastName },
    })
  },
  useHook: ({ fetch }) => () => {

    return useCallback(
      async function subscribe(input) {
        const data = await fetch({ input })
        return data
      },
      [fetch]
    )
  },
}
`


For my project I made types for everything extra so thats why they are imported but you can take those out I guess if you need this functionality urgently. Also you will need to add it to the provider files and commerce folder

Florian-crg:
What is in your @commerce/customer/use-subscribe ? Do you have a working exemple somewhere ?

kogan007:
> What is in your @commerce/customer/use-subscribe ? Do you have a working exemple somewhere ?
> 
> I wrote the code bellow, inspired from the use-signup.
> 
> ```
> import { useHook, useMutationHook } from '../utils/use-hook'
> import { mutationFetcher } from '../utils/default-fetcher'
> import type { HookFetcherFn, MutationHook } from '../utils/types'
> import type { SubscriptionsHook } from '../types/subscriptions'
[...]

PR #523, opened 2021-10-09
Add subscription endpoint for BC

Florian-crg:
Related issue https://github.com/vercel/commerce/issues/479


goncy:
Hey @Florian-crg, thanks but afaik we are not adding this to the core yet, we may reopen this in the future!

PR #527, opened 2021-10-14
make use responsive prefix

muhaimincs:


goncy:
Thanks @muhaimincs for the PR, but we are using directives in the rest of the code! Is this being deprecated or just a personal preference? Will close it on the meantime, thanks!

Issue #534, opened 2021-10-17
 CustomCheckout wrong button cart page


Florian-crg:
The button on the cart page `/cart`  is not the one that should be displayed when we enabled CustomCheckout.

```
next.config.js {
...
  "env": {
    "COMMERCE_CART_ENABLED": true,
    "COMMERCE_SEARCH_ENABLED": true,
    "COMMERCE_WISHLIST_ENABLED": true,
    "COMMERCE_CUSTOMERAUTH_ENABLED": true,
    "COMMERCE_CUSTOMCHECKOUT_ENABLED": true
  }
}
```

<img width="1192" alt="Capture d’écran 2021-10-17 à 21 42 51" src="https://user-images.githubusercontent.com/20863963/137642547-fcd12804-d7f3-4d62-b0e4-b98aebfeb712.png">

PR #535, opened 2021-10-17
Replace button custom checkout enabled on /cart

Florian-crg:
https://github.com/vercel/commerce/issues/534


goncy:
Thanks @Florian-crg !

computerjan:
Are you sure that works? `process.env` is not accessible in the browser, so the fix won't work.

PR #546, opened 2021-10-24
Custom checkout enhancements

cpv123:
## Background
I'm working on a [Commerce.js](https://commercejs.com/) provider with @robbieaverill which uses the new custom checkout feature. See the [PR here](https://github.com/vercel/commerce/pull/548). 

However, the existing structure for custom checkout is not very flexible making it difficult to implement properly. Specifically, it expects that cards and addresses can be separately submitted and associated with a customer before submitting checkout and does not allow an "all-in-one" checkout which includes card details, address details and order details.

## Description of changes
In general, this PR makes the card fields and address fields more accessible throughout the checkout UI and also available to the provider in certain checkout actions.

* Creates a `CheckoutContext` to store and update card and address fields
* Passes card and address fields to various checkout hooks allowing them to be used by the provider if necessary
* Not many changes to types, instead makes use of the existing `CheckoutTypes` type which has `checkout: any` property 
* Adds loading indicator when loading checkout submit
* Clears cart after successful checkout submit
* Adds an empty `useSubmitCheckout` hook into the `local` framework. **TODO** - does this need to be added for every existing provider because the build fails without it?

Note that there should be **no breaking changes** in this PR. Currently, Ordercloud is the only provider using the custom checkout and that should still work the same. Commerce.js is coming soon, and I will raise a draft PR for it soon which will show these changes in context.

## Final notes
Whilst these changes are to allow Commerce.js provider to use custom checkout, I think that the changes make custom checkout a lot more flexible in general and should benefit future providers. Based on the comment left here, it will immediately help Ordercloud: https://github.com/vercel/commerce/blob/main/framework/ordercloud/checkout/use-submit-checkout.tsx#L16

### DIFF
--- a/components/checkout/CheckoutSidebarView/CheckoutSidebarView.tsx
+++ b/components/checkout/CheckoutSidebarView/CheckoutSidebarView.tsx
@@ -1,5 +1,5 @@
 import Link from 'next/link'
-import { FC } from 'react'
+import { FC, useState } from 'react'
 import CartItem from '@components/cart/CartItem'
 import { Button, Text } from '@components/ui'
 import { useUI } from '@components/ui/context'
@@ -14,6 +14,8 @@ import PaymentWidget from '../PaymentWidget'
 import s from './CheckoutSidebarView.module.css'
 
 const CheckoutSidebarView: FC = () => {
+  const [loadingSubmit, setLoadingSubmit] = useState(false)
+
   const { setSidebarView, closeSidebar } = useUI()
   const { cardFields, addressFields, clearCheckoutFields } =
     useCheckoutContext()
@@ -27,18 +29,23 @@ const CheckoutSidebarView: FC = () => {
   const onSubmitCheckout = useSubmitCheckout()
 
   async function handleSubmit(event: React.ChangeEvent<HTMLFormElement>) {
-    event.preventDefault()
+    try {
+      event.preventDefault()
+      setLoadingSubmit(true)
 
-    await onSubmitCheckout({
-      checkout: {
-        addressFields,
-      },
-    })
-
-    // Assuming successful checkout, clear and close sidebar.
-    refreshCart()
-    clearCheckoutFields()
-    closeSidebar()
+      await onSubmitCheckout({
+        checkout: {
+          addressFields,
+        },
+      })
+      refreshCart()
+      clearCheckoutFields()
+      closeSidebar()
+    } catch {
+      // TODO - handle error UI here.
+    } finally {
+      setLoadingSubmit(false)
+    }
   }
 
   const { price: subTotal } = usePrice(
@@ -115,6 +122,7 @@ const CheckoutSidebarView: FC = () => {
             type="submit"
             width="100%"
             disabled={!checkoutData?.hasPayment || !checkoutData?.hasShipping}
+            loading={loadingSubmit}
           >
             Confirm Purchase
           </Button>
--- a/components/checkout/CheckoutSidebarView/CheckoutSidebarView.tsx
+++ b/components/checkout/CheckoutSidebarView/CheckoutSidebarView.tsx
@@ -35,6 +35,7 @@ const CheckoutSidebarView: FC = () => {
       },
     })
 
+    // Assuming successful checkout, clear and close sidebar.
     refreshCart()
     clearCheckoutFields()
     closeSidebar()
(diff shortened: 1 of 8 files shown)


goncy:
Hey @cpv123, thanks for the PR! I'm not fully convinced of having a context wrapping the entire app for storing those checkout fields, probably having just the sidebar being wrapped should be enough (I'm not looking at the code right now so idk if its possible). Also as we already have a `useCheckout` hook that handles operations and data from the checkout, having a `useCheckoutContext` doesn't feel to good to me, probably we can still create a context on the core with some methods to interact with a state and the `useContext` hook from the provider can consume that data and use it as needed.

tl;dr: I will take this info and will try to follow up with a solution that fits better our needs, or, if you think this message was clear enough and you want to try something like that, it would be great.

Also there are a few changes (a typo and a type) that would be nice to merge, if you want to create a PR for that, awesome! Otherwise I can add it on a new PR, just let me know.

Thanks for your help!

cpv123:
> Hey @cpv123, thanks for the PR! I'm not fully convinced of having a context wrapping the entire app for storing those checkout fields, probably having just the sidebar being wrapped should be enough (I'm not looking at the code right now so idk if its possible). Also as we already have a `useCheckout` hook that handles operations and data from the checkout, having a `useCheckoutContext` doesn't feel to good to me, probably we can still create a context on the core with some methods to interact with a state and the `useContext` hook from the provider can consume that data and use it as needed.
> 
> tl;dr: I will take this info and will try to follow up with a solution that fits better our needs, or, if you think this message was clear enough and you want to try something like that, it would be great.
> 
> Also there are a few changes (a typo and a type) that would be nice to merge, if you want to create a PR for that, awesome! Otherwise I can add it on a new PR, just let me know.
> 
> Thanks for your help!

@goncy thanks for taking a look at this.

1. Agreed that the context shouldn't wrap the entire app, but just the checkout part of the app (I think that I just copied the UI context without thinking properly)
2. I will pull out those small changes (the type and typo) and raise them in a separate PR
3. I think I understand what you're saying about the existing `useCheckout` hook - the reason I didn't use this was that I assumed that the hook must **fetch** the checkout data from the backend, whereas I needed access to the data (customer details/address/payment) before it had been submitted anywhere so that it could all be submitted at the end of the checkout. But I think you're saying that the `useCheckout` hook could do both? I'll take another look tomorrow with this perspective and see what I can do!

goncy:
> > Hey @cpv123, thanks for the PR! I'm not fully convinced of having a context wrapping the entire app for storing those checkout fields, probably having just the sidebar being wrapped should be enough (I'm not looking at the code right now so idk if its possible). Also as we already have a `useCheckout` hook that handles operations and data from the checkout, having a `useCheckoutContext` doesn't feel to good to me, probably we can still create a context on the core with some methods to interact with a state and the `useContext` hook from the provider can consume that data and use it as needed.
> > tl;dr: I will take this info and will try to follow up with a solution that fits better our needs, or, if you think this message was clear enough and you want to try something like that, it would be great.
> > Also there are a few changes (a typo and a type) that would be nice to merge, if you want to create a PR for that, awesome! Otherwise I can add it on a new PR, just let me know.
> > Thanks for your help!
> 
> @goncy thanks for taking a look at this.
> 
> 1. Agreed that the context shouldn't wrap the entire app, but just the checkout part of the app (I think that I just copied the UI context without thinking properly)
> 2. I will pull out those small changes (the type and typo) and raise them in a separate PR
> 3. I think I understand what you're saying about the existing `useCheckout` hook - the reason I didn't use this was that I assumed that the hook must **fetch** the checkout data from the backend, whereas I needed access to the data (customer details/address/payment) before it had been submitted anywhere so that it could all be submitted at the end of the checkout. But I think you're saying that the `useCheckout` hook could do both? I'll take another look tomorrow with this perspective and see what I can do!

Great about 1 and 2! You might be right with 3, I'm not looking at the code right now so I might not be full aware of the case, but I was thinking about storing the data in a context and using a hook (might not be core's `useCheckout`, I will have to take a deeper look to it), providers can access the information from it.

Still, this probably needs more thinking on our side to make it correct, so I will take a look at it with the team and let you know!

You can still use this solution for developing the provider and you can update it once we find a way to do it, hopefully it won't be much code on the provider side. If you can do the other small PR with the typos, that would be great.

Will close it by now but I will appreciate if you can create an issue explaining the same thing you added in here so we can keep track on it.

Again, thanks!

PR #548, opened 2021-10-27
Commerce.js Provider

cpv123:
### Changes to the main app
* Adds loading state to confirm checkout button and refreshes cart after custom checkout: [`0089819` (#548)](https://github.com/vercel/commerce/pull/548/commits/0089819607b57cf3f006a17f3dcc5bf28da06dbd)
* Two other very small commits to UI code: [`3609e86` (#548)](https://github.com/vercel/commerce/pull/548/commits/3609e867e407996e5c8ce3671c9a904bb13cebab) and [`f87a0b7` (#548)](https://github.com/vercel/commerce/pull/548/commits/f87a0b7a2b5bbd218546b976f75e363fb3187107)
* Allow the login API endpoint to handle GET requests and also URL paths: [`8ab961e` (#548)](https://github.com/vercel/commerce/pull/548/commits/8ab961e08ae3ac142443719a04dd83b3f9fc71e0) and [`c3713ec` (#548)](https://github.com/vercel/commerce/pull/548/commits/c3713ec6e23f1b423a071a31221069995d419486)

### New dependencies
1. Commerce.js SDK: [@chec/commerce.js](https://www.npmjs.com/package/@chec/commerce.js)
2. Types for SDK: [@types/chec__commerce.js](https://www.npmjs.com/package/@types/chec__commerce.js)

### Environment variables
* `NEXT_PUBLIC_COMMERCEJS_PUBLIC_KEY` - public key for a Commerce.js account.
* `NEXT_PUBLIC_COMMERCEJS_DEPLOYMENT_URL` - current deployment URL. This is optional but should be specified for "production deployments", otherwise the automatic `VERCEL_URL` variable will be used which might not be accurate when using a custom domain.

- [x]  CommerceProvider
- [ ]  Schema & TS types
- [ ]  API Operations - Get all collections
- [ ]  API Operations - Get all pages
- [x]  API Operations - Get all products
- [ ]  API Operations - Get page
- [x]  API Operations - Get product
- [x]  API Operations - Get Shop Info
- [x]  Hook - Add Item
- [x]  Hook - Remove Item
- [x]  Hook - Update Item
- [x]  Hook - Get Cart
- [x]  Auth
- [ ]  Customer information
- [x]  Product attributes - Size, Colors
- [x]  Custom checkout
- [x]  Typing
- [ ]  Tests


cpv123:
@goncy this PR currently relies on the custom checkout changes from https://github.com/vercel/commerce/pull/546 which we've since closed. I'm going to update this PR to try and get the checkout working even without those changes, so that hopefully this provider can be merged without any changes to the main codebase.

goncy:
Thanks @cpv123! Can you send me some credentials for a provisioned demo store at goncy@vercel.com ? I will start the review in the meantime

components/checkout/CheckoutSidebarView/CheckoutSidebarView.tsx
```diff
@@ -12,16 +12,24 @@ import PaymentWidget from '../PaymentWidget'
...
-    await onCheckout()
-
-    closeSidebar()
+      await onCheckout()
+      refreshCart()
+      closeSidebar()
+    } catch {
+      // TODO - handle error UI here.
+    } finally {
+      setLoadingSubmit(false)
```
goncy, line 31: Wouldn't setting this to false after closing the sidebar cause a warning for setting a state after the component has been unmounted?

README.md
```diff
@@ -13,6 +13,7 @@ Demo live at: [demo.vercel.store](https://demo.vercel.store/)
 - Vendure Demo: https://vendure.vercel.store
 - Saleor Demo: https://saleor.vercel.store/
 - Ordercloud Demo: https://ordercloud.vercel.store/
+- Commerce.js Demo: TODO
```
goncy, line 16: You can assume this would be commercejs.vercel.store

framework/commerce/new-provider.md
```diff
@@ -5,6 +5,7 @@ A commerce provider is a headless e-commerce platform that integrates with the [
 - BigCommerce ([framework/bigcommerce](../bigcommerce))
 - Saleor ([framework/saleor](../saleor))
 - Shopify ([framework/shopify](../shopify))
+- Commerce.js ([framework/shopify](../commercejs))
```
goncy, line 8: The url still contains shopify

framework/commercejs/.env.template
```diff
@@ -0,0 +1,7 @@
+COMMERCE_PROVIDER=commercejs
+
+# Public key for your Commerce.js account
+NEXT_PUBLIC_COMMERCEJS_PUBLIC_KEY=pk_test_347743e518b00d8165f453f2b099820ad9fd68a1fda91
```
goncy, line 4: The template contains the key

cpv123: 🤦‍♂️

components/checkout/CheckoutSidebarView/CheckoutSidebarView.tsx
cpv123: Good spot - my console warning confirms that you're correct. Fixed now

components/checkout/CheckoutSidebarView/CheckoutSidebarView.tsx
```diff
@@ -12,16 +12,23 @@ import PaymentWidget from '../PaymentWidget'
...
-    event.preventDefault()
+    try {
+      setLoadingSubmit(true)
+      event.preventDefault()
 
-    await onCheckout()
-
-    closeSidebar()
+      await onCheckout()
+      refreshCart()
```
goncy, line 26: We can ensure and call setLoadingSubmit to false in here? Just to be explicit

pages/api/login/index.ts
```diff
@@ -0,0 +1,4 @@
+import loginApi from '@framework/api/endpoints/login'
+import commerce from '@lib/api/commerce'
+
+export default loginApi(commerce)
```
goncy, line 4: Why was this page changed?

package.json
```diff
@@ -20,6 +20,7 @@
     "node": ">=14.x"
   },
   "dependencies": {
+    "@chec/commerce.js": "^2.8.0",
```
goncy, line 23: I realized that this sdk uses axios internally which is a pretty heavy lib, but its not being used a lot, probably using other fetch library might benefit the sdk also. This is not a hard requirement but its highly advised as currently all dependencies are installed for all providers (but just the ones needed are used)

pages/api/login/index.ts
cpv123: It happened as part of this commit: [`c3713ec` (#548)](https://github.com/vercel/commerce/pull/548/commits/c3713ec6e23f1b423a071a31221069995d419486). It's really just a rename from `login.ts --> login/index.ts`

Previously, there was a `pages/api/login.ts` file but I needed this API route to be dynamic, so I created a directory and moved:
```
pages/api/login.ts --> pages/api/login/index.ts
```
and added
```
pages/api/login/[...param].ts
```

I could have left it as `pages/api/login.ts` but I would have still created the directory for the dynamic route, so I thought it neater to put both routes within the directory (as `index.ts` and `[...param].ts`)

goncy: I'm not quite sure of adding this to the core, can you explain a little more on why this extra route is needed?

package.json
cpv123: The SDK is actually doing all the data fetching on both client and server-side, in the two "fetcher" functions which call SDK methods:

https://github.com/cpv123/commerce/blob/commercejs-provider/framework/commercejs/api/utils/sdk-fetch.ts
https://github.com/cpv123/commerce/blob/commercejs-provider/framework/commercejs/fetcher.ts

So I wouldn't want to remove the SDK dependency otherwise these fetcher functions would have to be re-written to make HTTP requests directly, and the benefits of using the SDK would be lost (such as standardised responses, errors, typings, etc.) Hopefully, this is OK for now.

There is a new version of the Commerce.js SDK being planned, and I'll make sure to leave some feedback about axios and the possibility of using a different (lighter) fetch library.

pages/api/login/index.ts
cpv123: **Background:**
Commerce.js only supports passwordless authentication - the user will enter their email address, and an email will be sent to them containing a "magic link" which will automatically authenticate them. This link's URL will contain a token, and when clicked, the link will take the user to an API that can exchange this token for a JWT and finally redirect the user to the application along with their JWT. 

**For example:**
1. User enters their email address and submits login form
2. User receives an email with link: `https://commercejs.vercal.store/api/login/{TOKEN}`
3. An API at route `/api/login/{TOKEN}` is called and exchanges `TOKEN` for a JWT, and redirects the user back to `https://commercejs.vercal.store` with JWT

**Why things needed to change:**
This authentication flow requires a login endpoint that accepts a dynamic path i.e. not just `/login` but `/login/{TOKEN}`.

The existing login API on its own will not satisfy this, because a request to `/login/123` will not match an API route at `/login`. This is why I created the new `/login/[...param]` API.

Let me know if any of that doesn't make sense!

goncy: Thanks for the heads up @cpv123 ! In that case I think you can add a rewrite on your providers next.config.js, something like this:

```js
const commerce = require('./commerce.config.json')

module.exports = {
  commerce,
  images: {
    domains: ['cdn.chec.io', 'ocdevops.blob.core.windows.net'],
  },
  rewrites() {
    return [{
      source: '/api/login/:token',
      destination: '/api/login?token=:token'
    }]
  }
}
```
And then on your `api/endpoints/login/login` you will receive a query param called token, that fits your needs?

package.json
ScopeyNZ: Commerce.js dev here! Yep the axios inclusion was a pretty old decision. We're planning to switch to fetch with the new version of Commerce.js.

pages/api/login/index.ts
cpv123: Amazing, thanks @goncy. 

I've reverted the original commit and instead used a rewrite as you've suggested.

goncy:
@cpv123 this PR adds a way to store checkout information to consume on other hooks https://github.com/vercel/commerce/pull/590 I will add some comments on those files as reference

framework/commercejs/checkout/use-submit-checkout.tsx
```diff
@@ -0,0 +1,30 @@
...
+    return data
+  },
+  useHook: ({ fetch }) =>
+    function useHook() {
+      return useCallback(async function onSubmitCheckout(input) {
+        const data = await fetch({ input })
+        return data
+      }, [])
+    },
+}
```
goncy, line 30:
```suggestion
import type { SubmitCheckoutHook } from '@commerce/types/checkout'
import type { MutationHook } from '@commerce/utils/types'

import { useCallback } from 'react'
import useSubmitCheckout, {
  UseSubmitCheckout,
} from '@commerce/checkout/use-submit-checkout'
import { useCheckoutContext } from '@components/checkout/context'

export default useSubmitCheckout as UseSubmitCheckout<typeof handler>

export const handler: MutationHook<SubmitCheckoutHook> = {
  fetchOptions: {
    url: '/api/checkout',
    method: 'POST',
  },
  async fetcher({ input: item, options, fetch }) {
    const data = await fetch({
      ...options,
      body: { item },
    })
    return data
  },
  useHook: ({ fetch }) =>
    function useHook() {
      const { cardFields, addressFields } = useCheckoutContext()

      return useCallback(async function onSubmitCheckout() {
        const data = await fetch({ input: {card: cardFields, address: addressFields} })
        return data
      }, [cardFields, addressFields])
    },
}
```

framework/commercejs/customer/card/use-add-item.tsx
```diff
@@ -0,0 +1,20 @@
...
+    url: '_',
+    method: '_',
+  },
+  useHook: () =>
+    function useHook() {
+      return useCallback(async function addItem() {
+        return undefined
+      }, [])
+    },
+}
```
goncy, line 20:
```suggestion
import type { AddItemHook } from '@commerce/types/customer/card'
import type { MutationHook } from '@commerce/utils/types'

import { useCallback } from 'react'
import useAddItem, { UseAddItem } from '@commerce/customer/card/use-add-item'
import { useCheckoutContext } from '@components/checkout/context'

export default useAddItem as UseAddItem<typeof handler>

export const handler: MutationHook<AddItemHook> = {
  fetchOptions: {
    url: '_',
    method: '_',
  },
  useHook: () =>
    function useHook() {
      const { setCardFields } = useCheckoutContext()

      return useCallback(async function addItem(input) {
        setCardFields(input)

        return undefined
      }, [setCardFields])
    },
}
```

goncy (changes requested):
I added some snippets for consuming the information from the context

framework/commercejs/README.md
```diff
@@ -0,0 +1,13 @@
+# [Commerce.js](https://commercejs.com/) Provider
+
+**Demo:** TODO
```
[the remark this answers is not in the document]
cpv123: Good catch, fixed 👍

framework/commercejs/fetcher.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!resource) {
+    throw new FetcherError({
+      errors: [
+        { message: `Query ${query} does not exist on Commerce.js SDK.` },
+      ],
+      status: 400,
+    })
+  }
+
+  // @ts-ignore
```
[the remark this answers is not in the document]
cpv123: @goncy I've fixed the @ts-ignores and TODOs within the Commerce.js provider, except for the TODO regarding the checkout issue which you've since fixed. I'm going to work on that now, along with your code suggestions (thanks!)

components/checkout/CheckoutSidebarView/CheckoutSidebarView.tsx
```diff
@@ -10,18 +10,29 @@ import useCheckout from '@framework/checkout/use-checkout'
...
-    event.preventDefault()
+    try {
+      setLoadingSubmit(true)
+      event.preventDefault()
 
-    await onCheckout()
-
-    closeSidebar()
+      await onCheckout()
+      clearCheckoutFields()
```
cpv123, line 28: @goncy I added this which clears the checkout context after a successful checkout.

I was wondering if the `clearCheckoutFields` function should be returned from the `useCheckout` hook directly (like the `submit` function) so that we never have to interact with **both** that and `useCheckoutContext`.

But I decided not to go for this because I noticed that it would probably have to be returned as a "mutation" according to the type (like `submit` currently is) which isn't exactly correct semantically. 

```
export type GetCheckoutHook<T extends CheckoutTypes = CheckoutTypes> = {
  data: T['checkout'] | null
  input: {}
  fetcherInput: { cartId?: string }
  swrState: { isEmpty: boolean }
  mutations: { submit: UseSubmitCheckout }   <--- could add `clearCheckoutFields` here
}
```

Let me know what you think.

goncy: Clearing the checkout fields there seems ok, mutations usually mutates or interacts with the data it manages, in this case is just a side effect, I preffer to leave it in its own place but we can get back at it later if it starts to feel weird

cpv123: @goncy anything else to address in this PR?

goncy: Hey @cpv123 can you please resolve the current conflicts? Thanks!

goncy:
Hey @cpv123 we are almost there! I edited the PR comment with a checklist, not all of them are required so you can fill what the provider covers

framework/commercejs/customer/use-customer.tsx
```diff
@@ -0,0 +1,44 @@
+import Cookies from 'js-cookie'
+import { decode } from 'jsonwebtoken'
+import { SWRHook } from '@commerce/utils/types'
+import useCustomer, { UseCustomer } from '@commerce/customer/use-customer'
+import { CUSTOMER_COOKIE, API_URL } from '@framework/constants'
```
goncy, line 5: We shouldn't import from framework inside the provider as it might break other providers

goncy (changes requested):
Some small changes

goncy:
@cpv123 can you solve the conflicts? It should be really small documentation changes

cpv123:
> @cpv123 can you solve the conflicts? It should be really small documentation changes

@goncy conflicts are resolved, and checklist items are ticked.

goncy (approved):

goncy:
Merged and live at https://commercejs.vercel.store/ 🎉, congrats and thanks for the PR!

PR #562, opened 2021-11-09
Update Next.js to 12 and switch to npm

lfades:
I tried keeping both `yarn.lock` and `package-lock.json` but doing a npm install always created changes in `yarn.lock`.

Other changes are related to the breaking change introduced by `next/image` where the container changed from a `div` to `span`.


charkour:
Out of curiosity, what's the reason of switching from Yarn to npm? (I switched to npm after forking the project myself)

lfades:
@charkour I tried to keep both, as the project should work with both, but to avoid warnings and because every time I installed packages with npm it changed the entire yarn.lock, I decided to stick to one, and npm is more beginner friendly as it comes pre-installed with Node.js

PR #563, opened 2021-11-09
Abstract common static props

luisorbaiceta:
I have added a wrapper for the `getStaticProps` function that adds all the pages common props and expands the `context` object with those defaults.

## Goal
The main purpose of this PR is to centralize the common props handling in order to ensure a correct implementation across all the project.

- [x] fix some types of the wrapper

### DIFF
--- a/lib/default-props.ts
+++ b/lib/default-props.ts
@@ -7,6 +7,7 @@ import type { Page } from '@commerce/types/page'
 import type { Category } from '@commerce/types/site'
 
 import commerce from '@lib/api/commerce'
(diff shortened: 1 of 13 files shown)


luisorbaiceta:
Any thoughts about this @lfades ?

goncy:
Hey @luisorbaiceta, we have been talking about this with the team, we tried some abstractions before but nothing that totally fit our needs right now, we will continue iterating until found something, so I will close this by now, please continue with the great ideas and contributions, it creates great conversations about how we plan our future work, thanks!

PR #570, opened 2021-11-14
Add options back

Florian-crg:


goncy:
@Florian-crg is this still valid?

PR #574, opened 2021-11-17
Fix variable name

greyhere:
Variable name should be `ChevronRight`


goncy:
You are ChevronRight! Thanks 🙌

goncy (approved):

PR #577, opened 2021-11-23
Change "forgot" to "forget" in LoginView account recovery link

tniezg:


goncy:
@tniezg Afaik this looks correct, I attach an example from gmail
image

PR #578, opened 2021-11-23
Make wishlist's useRemoveItem call in WishlistCard consistent with WishlistButton

tniezg:
In `WishlistCard`, the remove wishlist item hook is called like so: https://github.com/vercel/commerce/blob/main/components/wishlist/WishlistCard/WishlistCard.tsx#L43. *The `id` value contains the id of the product.*

This is how the remove hook is called in `WishlistButton`: https://github.com/vercel/commerce/blob/main/components/wishlist/WishlistButton/WishlistButton.tsx#L53. *The `id` value is the wishlist item's id.*

While the `Wishlist` type is currently `any` (https://github.com/vercel/commerce/blob/main/framework/commerce/types/wishlist.ts#L2), I think the best approach is to agree the value of `id` to be the id of the wishlist item and not the product it's associated with. It's how it already works in `WishlistButton` and in https://github.com/vercel/commerce/blob/main/components/wishlist/WishlistButton/WishlistButton.tsx#L34 a separate `product_id` field is referenced.

The above means changing how `WishlistCard` calls the remove hook. The PR contains this change.

### DIFF
--- a/components/wishlist/WishlistCard/WishlistCard.tsx
+++ b/components/wishlist/WishlistCard/WishlistCard.tsx
@@ -11,14 +11,16 @@ import type { Product } from '@commerce/types/product'
 import usePrice from '@framework/product/use-price'
 import useAddItem from '@framework/cart/use-add-item'
 import useRemoveItem from '@framework/wishlist/use-remove-item'
+import { Wishlist } from '@commerce/types/wishlist'
 
 interface Props {
-  product: Product
+  item: Wishlist
 }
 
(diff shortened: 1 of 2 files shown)


lfades (approved):

goncy:
Great @tniezg, thanks! I will follow up with the provider PR 🙌, can you sync that one?

tniezg:
Thanks for the CR.

@goncy 👋 Synced.

PR #588, opened 2021-11-30
Allow cart to handle products without images

cpv123:
Previously, products without images would break the cart sidebar view because the `<Image />` component would not receive a `src`.

Now, it'll only render the `<Image />` is a `src` exists.


components/cart/CartItem/CartItem.tsx
```diff
@@ -86,15 +86,17 @@ const CartItem = ({
...
-              <Image
-                onClick={() => closeSidebarIfPresent()}
-                className={s.productImage}
-                width={150}
-                height={150}
-                src={item.variant.image!.url}
-                alt={item.variant.image!.altText}
-                unoptimized
-              />
+              {item?.variant?.image?.url && (
```
goncy, line 89: Shouldn't this be wrapping the entire link element? Otherwise we have an empty link

cpv123: @goncy I've updated this now to match the behaviour of product cards, where if an image doesn't exist it will show a placeholder image. This way, the link is always available even when there is no image.

Screenshot 2564-12-04 at 17 23 59

components/cart/CartItem/CartItem.tsx
```diff
@@ -1,15 +1,15 @@
...
 import Link from 'next/link'
 import s from './CartItem.module.css'
-import { Trash, Plus, Minus, Cross } from '@components/icons'
 import { useUI } from '@components/ui/context'
 import type { LineItem } from '@commerce/types/cart'
 import usePrice from '@framework/product/use-price'
 import useUpdateItem from '@framework/cart/use-update-item'
 import useRemoveItem from '@framework/cart/use-remove-item'
 import Quantity from '@components/ui/Quantity'
+import { placeholderImg } from '@components/product/ProductCard'
```
goncy, line 12: I think is better to just duplicate the url as this creates a dependency between ProductCard and CartItem that doesn't really exist, we can later move this url to some constants file and import it from there

goncy (changes requested):

goncy:
Hey @cpv123 ! I will close this to avoid getting conflicts, but I already merged a fix for it, thanks 🙌

PR #594, opened 2021-12-07
Fix BigCommerce Checkout SSO redirect_to issue

jordanarldt:
### Issue
When using the BigCommerce integration, Customer Logins are not maintained when proceeding to the redirected checkout. This is because the **redirect_to** parameter when generating the Customer JWT is using an absolute URL when the API only allows relative paths. 

### Fix
In **framework/bigcommerce/api/endpoints/checkout/get-checkout.ts** on line 45, the **redirect_to** parameter is this currently:
```tsx
redirect_to: data.checkout_url,
```
But needs to be replaced with this:
```tsx
redirect_to: data.checkout_url.replace(config.storeUrl, ""),
```

After making this change, the BigCommerce redirected checkout maintains the customer login session.


goncy:
Hey @jordanarldt, thanks for the PR, I'm getting this error when doing checkout
image

Do you know what might be causing it?

goncy:
Nevermind, seems working with the deploy url but not with the one generated in the PR

PR #606, opened 2021-12-14
Remove unnecessary check in the Sidebar component

lsalling:
The cleanup function for the Sidebar component tries to remove a scroll lock of an element that, by the time the code runs, has most likely already been removed (which is also stated by a lint warning):

`
./components/common/Sidebar/Sidebar.tsx
31:59  Warning: The ref value 'contentRef.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'contentRef.current' to a variable inside the effect, and use that variable in the cleanup function.
react-hooks/exhaustive-deps
`

All scroll locks are removed in the proceeding `clearAllBodyScrollLocks` call either way.


lsalling:
I see the same thing happens in the [Modal](https://github.com/vercel/commerce/blob/main/components/ui/Modal/Modal.tsx) base though this time being saved in a variable inside the useEffect function.
Is there a particular reason as to this?

goncy:
Hey @lsalling I have checked the docs of body-scroll-lock and it looks like just calling `clearrAllBodyScrollLocks` should be enough, can you remove this cases in other files if present and do a quick pass on your side to see if it works correctly? I will also do it on my side, thanks!

lsalling:
Hi - it's now corrected in the Modal component as well and did not find other occurrences. Checked on local and everything worked as supposed to! :slightly_smiling_face:

goncy (approved):

goncy:
Thanks @lsalling !

Issue #614, opened 2021-12-16
Having issue with next/dynamic


geniti:
![Schermafbeelding 2021-12-16 om 10 09 47](https://user-images.githubusercontent.com/32541489/146342045-f26e028a-1492-430d-bbb4-8c6269c8a9d0.png)

geniti:
@lfades do you know how to solve this err?

lsalling:
There's a problem when installing dependencies using yarn right now. Use npm instead🙂

Dooptroop:
You can get it working with yarn by updating the `Layout.tsx` files dynamic imports like so

```typescript
const Loading = () => (
  <div className="w-80 h-80 flex items-center text-center justify-center p-3">
    <LoadingDots />
  </div>
)

const SignUpView = dynamic(
  () => import('@components/auth/SignUpView'),
  {
    loading: Loading
  }
)

const ForgotPassword = dynamic(
  () => import('@components/auth/ForgotPassword'),
  {
    loading: Loading
  }
)

const FeatureBar = dynamic(
  () => import('@components/common/FeatureBar'),
  {
    loading: Loading
  }
)

const Modal = dynamic(
  () => import('@components/ui/Modal'),
  {
    loading: Loading,
    ssr: false
  }
)
```

Dooptroop:
There is a package-lock.json commited so npm install maintains the package versions.

Running with Yarn will install slightly newer versions of the packages.

Dooptroop:
https://github.com/vercel/commerce/pull/618

Dooptroop:
I made a PR here https://github.com/vercel/commerce/pull/618

PR #618, opened 2021-12-17
Fixes issue with next/dynamic when running next version > 12.0.3

Dooptroop:
Fixes issue seen with next/dynamic when running yarn install due to problems with newer versions of next

Regarding issue #614 

https://github.com/vercel/next.js/pull/31683

### DIFF
--- a/components/common/Layout/Layout.tsx
+++ b/components/common/Layout/Layout.tsx
@@ -25,31 +25,35 @@ const Loading = () => (
   </div>
 )
 
+const dynamicProps = {
+  loading: Loading,
+}
+
 const SignUpView = dynamic(
   () => import('@components/auth/SignUpView'),
   {
-    loading: Loading
+    ...dynamicProps
   }
 )
 
 const ForgotPassword = dynamic(
   () => import('@components/auth/ForgotPassword'),
   {
-    loading: Loading
+    ...dynamicProps
   }
 )
 
 const FeatureBar = dynamic(
   () => import('@components/common/FeatureBar'),
   {
-    loading: Loading
+    ...dynamicProps
   }
 )
 
 const Modal = dynamic(
   () => import('@components/ui/Modal'),
   {
-    loading: Loading,
+    ...dynamicProps,
     ssr: false
   }
 )
--- a/components/common/Layout/Layout.tsx
+++ b/components/common/Layout/Layout.tsx
@@ -25,28 +25,33 @@ const Loading = () => (
   </div>
 )
 
-const dynamicProps = {
-  loading: Loading,
-}
-
 const SignUpView = dynamic(
   () => import('@components/auth/SignUpView'),
-  dynamicProps
+  {
+    loading: Loading
+  }
 )
 
 const ForgotPassword = dynamic(
   () => import('@components/auth/ForgotPassword'),
-  dynamicProps
+  {
+    loading: Loading
+  }
 )
 
 const FeatureBar = dynamic(
   () => import('@components/common/FeatureBar'),
-  dynamicProps
+  {
+    loading: Loading
+  }
 )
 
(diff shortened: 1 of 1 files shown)


components/common/Layout/Layout.tsx
```diff
@@ -25,28 +25,33 @@ const Loading = () => (
   </div>
 )
 
-const dynamicProps = {
-  loading: Loading,
```
ancaemcken, line 29: You should keep this assignment, then use the spread syntax `{ ...dynamicProps}` instead of `dynamicProps`. 

That way, you can have common `dynamicProps` without repeating yourself if you need to add some other properties than `loading`. 

It makes the solution easier to maintain.

components/common/Layout/Layout.tsx
```diff
@@ -25,28 +25,33 @@ const Loading = () => (
...
-  dynamicProps
+  {
+    loading: Loading
+  }
 )
 
 const ForgotPassword = dynamic(
   () => import('@components/auth/ForgotPassword'),
-  dynamicProps
+  {
```
ancaemcken, line 37: Revert `dynamicProps` and in here use `{...dynamicProps}`.

components/common/Layout/Layout.tsx
```diff
@@ -25,28 +25,33 @@ const Loading = () => (
...
+    loading: Loading
+  }
 )
 
 const FeatureBar = dynamic(
   () => import('@components/common/FeatureBar'),
-  dynamicProps
+  {
+    loading: Loading
```
ancaemcken, line 45: Revert `dynamicProps` and in here use `{...dynamicProps}`.

components/common/Layout/Layout.tsx
```diff
@@ -25,28 +25,33 @@ const Loading = () => (
...
+  }
 )
 
 const Modal = dynamic(
   () => import('@components/ui/Modal'),
-  Object.assign(dynamicProps, { ssr: false })
+  {
```
ancaemcken, line 51: Revert `dynamicProps` and in here replace this object with: 
```
{
  ...dynamicProps,
  ssr: false
}
```

components/common/Layout/Layout.tsx
```diff
@@ -25,28 +25,33 @@ const Loading = () => (
...
 
-const dynamicProps = {
-  loading: Loading,
-}
-
 const SignUpView = dynamic(
   () => import('@components/auth/SignUpView'),
-  dynamicProps
+  {
+    loading: Loading
```
ancaemcken, line 31: Revert `dynamicProps` and in here use {...dynamicProps}.

ancaemcken (changes requested):

components/common/Layout/Layout.tsx
Dooptroop: Yes, that's smarter. :+1:

Dooptroop:
Updated to use prop spreading as suggested by @ancaemcken

ancaemcken (approved):

components/common/Layout/Layout.tsx
```diff
@@ -31,22 +31,31 @@ const dynamicProps = {
 
 const SignUpView = dynamic(
   () => import('@components/auth/SignUpView'),
-  dynamicProps
+  {
+    ...dynamicProps
```
goncy, line 35: For the cases that we don't need to extend the object can you leave it like before? Or does it have any problem with it?

goncy (changes requested):

components/common/Layout/Layout.tsx
dylf: `dynamic()` needs an object literal, the way it was before was causing the issue

PR #660, opened 2022-01-26
feat: replace next-seo with custom solution

dominiksipowicz:
## Why

- Reduce the size
- Bundlephobia link: https://bundlephobia.com/package/next-seo@4.29.0
- We would like to simplify and extend default settings while customizing SEO

## Considerations
- Multiple open graph images were unable to be properly tested, as the social media platforms I tested could not properly select from multiple images.
- changed `seo.json` to `seo_meta.json` to accommodate JSON+LD data structure files in the future.


components/common/Head/Head.tsx
```diff
@@ -1,17 +1,16 @@
 import { FC } from 'react'
-import NextHead from 'next/head'
-import { DefaultSeo } from 'next-seo'
-import config from '@config/seo.json'
+import { SEO } from '@components/common'
```
gbibeaul, line 2: could add import type to get that remove in the built version

gbibeaul (dismissed):
Left a tiny knitpicking comment.

components/common/SEO/SEO.tsx
```diff
@@ -0,0 +1,129 @@
...
+        key="og:url"
+        property="og:url"
+        content={openGraph?.url ?? config.openGraph.url}
+      ></meta>
+      {openGraph?.locale && (
+        <meta key="og:locale" property="og:locale" content={openGraph.locale} />
+      )}
+      {openGraph?.images?.length
+        ? openGraph.images.map((img, index) => ogImage(img, index))
+        : ogImage(config.openGraph.images[0], 0)}
```
goncy, line 97: I think that this should be a component instead of being called as a function

config/seo_meta.json
```diff
@@ -6,14 +6,13 @@
...
-    "locale": "en_IE",
-    "url": "https://nextjs.org/commerce",
+    "url": "https://demo.vercel.store/",
     "site_name": "Next.js Commerce",
     "images": [
       {
-        "url": "/card.png",
-        "width": 800,
-        "height": 600,
+        "url": "https://demo.vercel.store/card.png",
```
goncy, line 13: This points only to the bigcommerce version, is that expected?

components/common/SEO/SEO.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <meta
+        key={`og:image:alt${index}`}
+        property="og:image:alt"
+        content={alt}
+      />
+    </>
+  )
+}
+
+const SEO: FC<Props> = ({ title, description, openGraph, robots }) => {
```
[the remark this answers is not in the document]
dominiksipowicz: nice catch, I actually wanted to use `children` but forgot to render them :

config/seo_meta.json
dominiksipowicz: good point, I added new ENV var

components/common/SEO/SEO.tsx
dominiksipowicz: I tried to do that as a JSX component but either it was not updating properly meta tags between the rerenders (https://github.com/vercel/commerce/pull/660/files#diff-ae719305395f324d42d061ca30f7e380b3383e94ccda8ce4b85dcfe7f4ed31dcR70) or I could return an array of `ReactNode[]` then TypeScript was complaining and failing build.

site/components/common/SEO/SEO.tsx
```diff
@@ -0,0 +1,158 @@
+import Head from 'next/head'
+import { FC, Fragment, ReactNode } from 'react'
+import config from '@config/seo_meta.json'
+
+const storeBaseUrl =
+  process.env.NEXT_PUBLIC_STORE_URL ?? process.env.VERCEL_URL
+    ? `https://${process.env.VERCEL_URL}`
+    : null
```
lfades, line 8: I don't think this is going to work, `VERCEL_URL` won't be replaced by Next.js at build time. It needs to be added to `next.config.js` so that it gets replaced like variables that start with `NEXT_PUBLIC_`

site/components/common/SEO/SEO.tsx
```diff
@@ -0,0 +1,158 @@
...
+  title,
+  description,
+  openGraph,
+  robots,
+  children,
+}) => {
+  /**
+   * @see https://nextjs.org/docs/api-reference/next/head
+   *
+   * meta or any other elementsneed to be contained as direct children of the Head element,
```
lfades, line 72:
```suggestion
   * meta or any other elements need to be contained as direct children of the Head element,
```

lfades (commented):
Added some comments

goncy (approved):

lfades (approved):
Thank you!

aecorredor:
@dominiksipowicz just found this PR while checking out how to add more metadata for instagram + facebook. I saw you guys added Twitter here. I was wondering, are there any guidelines for adding instagram and facebook? I'd be happy to try to put something up if you point me in the right direction. Thanks!

CraigWatt:
> Added new optional ENV variable NEXT_PUBLIC_STORE_URL which makes OG image URL absolute instead of relative. Some platforms (eg. Twitter) render only OG images when a full URL is present.
fallback to NEXT_PUBLIC_VERCEL_URL is used for Vercel deployments.

Hi @dominiksipowicz , these variable names are not included in the .env.template on main branch.  

Should they be to help avoid confusion?

dominiksipowicz:
> @dominiksipowicz just found this PR while checking out how to add more metadata for instagram + facebook. I saw you guys added Twitter here. I was wondering, are there any guidelines for adding instagram and facebook? I'd be happy to try to put something up if you point me in the right direction. Thanks!

Hi @aecorredor, The purpose of this PR was to minimise the dependencies for the commerce project and give developers more freedom to customize. If you have a lightweight implementation please submit a PR. You can use this PR as an example how to add more OG tags.

dominiksipowicz:
> > Added new optional ENV variable NEXT_PUBLIC_STORE_URL which makes OG image URL absolute instead of relative. Some platforms (eg. Twitter) render only OG images when a full URL is present.
> > fallback to NEXT_PUBLIC_VERCEL_URL is used for Vercel deployments.
> 
> Hi @dominiksipowicz , these variable names are not included in the .env.template on main branch.
> 
> Should they be to help avoid confusion?

@CraigWatt I think the `NEXT_PUBLIC_STORE_URL` can be added. Good spot! Thanks. Do you want to submit PR?

PR #671, opened 2022-02-07
To implement checkout tax and address

saravanan-trika:
- Checkout tax implementation
- customer address stored to elastic-path


saravanan-trika:
Wrong pull request

PR #679, opened 2022-02-11
Fix turborepo caching edge case 

lfades:
The cache currently depends on the `COMMERCE_PROVIDER` env var, however I went through the import flow of Next.js Commerce again and it failed because our org already has the build cached, and that's okay but because the provider env isn't set then it's getting the output of the `local` provider. With this PR the cache now includes all the envs we use for package detection when `COMMERCE_PROVIDER` isn't set.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -6,7 +6,13 @@
       "outputs": ["dist/**"]
     },
     "next-commerce#build": {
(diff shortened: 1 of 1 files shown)


dominiksipowicz (approved):
lgtm

PR #691, opened 2022-02-20
Add lodash.debounce to Shopify provider package.json dependencies

lkostrowski:
Shopify provider requires `lodash.debounce` package. This package is installed in BigCommerce provider, but not in Shopify provider. After BigCommerce is removed, Shopfify provider can't be installed.

This PR fixes this error

### DIFF
--- a/packages/commercejs/package.json
+++ b/packages/commercejs/package.json
@@ -48,7 +48,8 @@
   },
   "dependencies": {
     "@chec/commerce.js": "^2.8.0",
-    "@vercel/commerce": "^0.0.1"
+    "@vercel/commerce": "^0.0.1",
+    "lodash.debounce": "^4.0.8"
   },
   "peerDependencies": {
     "next": "^12",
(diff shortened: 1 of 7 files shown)


okbel (approved):

okbel:
It's been used by multiple providers. Can you add it to site, instead of just the shopify provider?

okbel (changes requested):
Most providers need debounce. Can you add it to site? Thank you!

lkostrowski:
@okbel are you sure about that? It doesn't seem like a good way to manage dependencies. Its provider(s) that needs this dep, not client. I suggest adding it to every provider that explicitly needs it, instead of the client implicitly.

okbel:
If it's not used in the UI (core and site) and it's specific for the provider then let's add it to all of those who are using it.

lkostrowski:
@okbel done, added to all providers that use debounce (few of them don't)

okbel (approved):

Issue #688, opened 2022-02-17
Error when deploying on Vercel Bigcommerce


GodwinEbikwo:
After signing up and connecting the store, the build failed with the following error:

Error: No Output Directory named "public" found after the Build completed. You can configure the Output Directory in your Project Settings. Learn More: https://vercel.link/missing-public-directory

@lfades

omertoast:
I'm also having the same problem

cpv123:
@GodwinEbikwo I got this working by configuring some settings in Vercel. I've explained here: https://github.com/vercel/commerce/pull/692

lfades:
Should be working now!

https://github.com/vercel/commerce/pull/692#issuecomment-1055863907

Issue #672, opened 2022-02-07
Deploying new instance of Nextjs Commerce in Vercel fails with 404


robearlam:
Hi,

I've attempted several times to deploy a new instance of Nextjs Commerce through the Vercel portal, but it fails. The build output looks to be very truncated compared with previous successful deployments I've done in the past.

```
07:56:37.771 | Cloning completed: 798.906ms
07:56:38.035 | Analyzing source code...
07:56:39.501 | Uploading build outputs...
07:56:45.681 | Deploying build outputs...
07:56:55.524 | Done with "."
```

When accessing the deployed site, I am presented with a 404 error:
![image](https://user-images.githubusercontent.com/2149251/152870964-f8549a48-7677-46d4-ae00-358d893f8eae.png)

You can see the erroring instance I'm trying to access here: https://commerce-fhgww6ugq-robearlam.vercel.app/
The repo used to deploy from is: https://github.com/robearlam/commerce (This is the default cloned code, without any changes)

Thanks!

juancamiloqhz:
I'm having the same issue 
Change the build comman to 
cd ../ && npx turbo run build --scope=site --include-dependencies --no-deps
And also change project folder to "site"
Then manually trigger a redeploy to build the app, but still 404 after build is complete

neneji:
what i did is:
in `Build & Development Settings`
1. set the `FRAMEWORK PRESET` to `Next.js`
2. set the `BUILD COMMAND` to `turbo run build`
2. set the `OUTPUT DIRECTORY` to `site/.next`

i dont know if there is a better way to config it or not, cannot find any guide about that...

dominiksipowicz:
Hi @robearlam @juancamiloqhz 
Please try using the link https://vercel.com/new and selecting "Next.js Commerce". It should automatically clone the repository to your GitHub account, choose the integration (or skip provider) and deploy to Vercel automatically with the correct settings.

robearlam:
This was related to the change over to monorepo, looks to have been solved now.

guibovo:
I've forked the repo - worked a bit on it and deployed on Vercel, I'm still facing this issue.

https://vercel.com/guibovo/commerce/Dmdu8ExwiZxwqzcUK2WyroQiA2xL

<img width="961" alt="image" src="https://user-images.githubusercontent.com/3513557/183699996-ee651a4f-8bfb-4714-b38f-adc424af0a35.png">



[...]

PR #692, opened 2022-02-21
Add README instructions for Vercel deployment

cpv123:
Since migrating to Turborepo, it seems that deploying to Vercel requires some "Build and Output" configuration extra steps but these aren't documented anywhere. This is what worked for me.

There are a few issues already:
https://github.com/vercel/commerce/issues/688
https://github.com/vercel/commerce/issues/672

Screenshot of my Vercel dashboard (which is now working):
Screenshot 2565-02-21 at 16 54 18

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,12 @@ Follow our docs for [Adding a new Commerce Provider](packages/commerce/new-provi
 
 If you succeeded building a provider, submit a PR with a valid demo and we'll review it asap.
 
+## Deploying to Vercel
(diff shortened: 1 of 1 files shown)


lfades:
Hi @cpv123 👋 

We were aware of this issue but instead of fixing it here I did some changes to the deploy button flow in Vercel in order to add those settings from the deploy button instead (it's live).

Thank you for the PR and let me know if it still fails for you!

PR #695, opened 2022-02-23
add FAQ for cannot find module error

dominiksipowicz:
Preview: https://github.com/vercel/commerce/tree/faq-cannot-find-module-error#troubleshoot


lfades:
@devnajam It's not exactly an issue, the app relies on external packages that need to be built for it to work, and they're built when you run the app from the root (using the monorepo)

lfades (approved):
LGTM. Thank you!

PR #709, opened 2022-03-12
Fix navigation between cart and menu on mobile

ltueth:
On mobile when the Sidebar menu is open and you want to switch to cart, it would toggle the sidebar and close it instead of switching view.


okbel (dismissed):

HeavenlyEntity:
Your menu items on mobile do not render on local nor Shopify I would advise using an "effect" to get them to render.

GuiBibeau (approved):
Thanks!

Issue #682, opened 2022-02-14
Add to cart throwing 500 error


brightonmike:
Just set up a new deploy - everything fine except add to cart, which throws a function error.

```
2022-02-14T17:26:18.526Z	f3136d01-4b70-45e4-9442-0c7918c1fbb3	ERROR	Error: Cannot find module 'immutability-helper'
Require stack:
- /var/task/site/.next/server/pages/api/cart.js
- /var/task/node_modules/next/dist/server/base-server.js
- /var/task/node_modules/next/dist/server/next-server.js
- /var/task/___next_launcher.js
- /var/runtime/UserFunction.js
- /var/runtime/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:93:18)
    at Object.6333 (/var/task/site/.next/server/pages/api/cart.js:25:18)
    at __webpack_require__ (/var/task/site/.next/server/webpack-api-runtime.js:25:42)
    at Object.5360 (/var/task/site/.next/server/chunks/294.js:767:37)
    at __webpack_require__ (/var/task/site/.next/server/webpack-api-runtime.js:25:42)
    at Object.2411 (/var/task/site/.next/server/pages/api/cart.js:121:17)
    at __webpack_require__ (/var/task/site/.next/server/webpack-api-runtime.js:25:42) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/var/task/site/.next/server/pages/api/cart.js',
    '/var/task/node_modules/next/dist/server/base-server.js',
    '/var/task/node_modules/next/dist/server/next-server.js',
    '/var/task/___next_launcher.js',
    '/var/runtime/UserFunction.js',
    '/var/runtime/index.js'
  ]
}
RequestId: f3136d01-4b70-45e4-9442-0c7918c1fbb3 Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

okbel:
Is this still happening?

PR #719, opened 2022-03-23
Bug Fix Round and Updates

### LINKED ISSUE #468: 404 page is not loading
`pages/404.tsx` is not being picked up by Next.js - https://github.com/vercel/commerce/blob/main/pages/404.tsx

This file has the text ` The requested page doesn't exist or you don't have access to it.`

*But* the 404 page looks like the default Next.js one e.g. https://demo.vercel.store/foo-bar shows `This page could not be found.`

This means we can't set our own custom 404 page :smiley:

okbel:
- This PR adds a missing dependency for Kibo Commerce https://github.com/vercel/commerce/issues/682
- removed dead and unused code
- refactored how we handle missing pages. It's now returning 404 instead of throwing on the build. This is something we could discuss because there are valid 404 (even on build) and there are definitely worrying situations when something should be and is not present. This will fix #468
- refactored types for pages. I get flaky errors on a few builds around those types returning `never`.

```
next-commerce:build: (node:919) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'pages' of undefined
next-commerce:build: TypeError: Cannot read property 'pages' of undefined
next-commerce:build: (node:926) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'collections' of undefined
```


dominiksipowicz (approved):

dominiksipowicz (approved):

ashishkpaul:
Hello sir in spree/commerce  its not fetching the products from menu links. is there any solution
i tried app link https://commerce-spree-git-kib-deps-vercel-solutions-vtest314.vercel.app/
and also on the fresh deployment showing the same results

spree-vercel

Issue #681, opened 2022-02-12
Error: Cannot find module '@vercel\commerce-local\next.config'


radgezito:
After cloning the project, creating the .env.local file to **use @vercel/commerce-local** I get an error message when I run **yarn dev** command

$ yarn dev
yarn run v1.22.10
$ next dev
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
**info  - Loaded env from c:\test\commerce\site\.env.local
error - Failed to load next.config.js, see more info here https://nextjs.org/docs/messages/next-config-error
Error: Cannot find module '@vercel\commerce-local\next.config'**
Require stack:
- c:\test\commerce\site\noop.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.mod._resolveFilename (c:\test\commerce\node_modules\next\dist\build\webpack\require-hook.js:179:28)
    at resolveFileName (c:\test\commerce\node_modules\resolve-from\index.js:29:39)
    at resolveFrom (c:\test\commerce\node_modules\resolve-from\index.js:43:9)
    at module.exports (c:\test\commerce\node_modules\resolve-from\index.js:46:47)
    at module.exports (c:\test\commerce\node_modules\import-from\index.js:4:55)
    at module.exports (c:\test\commerce\node_modules\import-cwd\index.js:4:30)
    at Object.withCommerceConfig (c:\test\commerce\packages\commerce\dist\config.cjs:19:30)
    at withCommerceConfig (c:\test\commerce\site\commerce-config.js:102:15)
    at Object.<anonymous> (c:\test\commerce\site\next.config.js:11:18) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'C:\\temp\\delete\\commerce\\site\\noop.js' ]
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

amir-zahedi:
> After cloning the project, creating the .env.local file to **use @vercel/commerce-local** I get an error message when I run **yarn dev** command
> 
> $ yarn dev yarn run v1.22.10 $ next dev ready - started server on 0.0.0.0:3000, url: http://localhost:3000 **info - Loaded env from c:\test\commerce\site.env.local error - Failed to load next.config.js, see more info here https://nextjs.org/docs/messages/next-config-error Error: Cannot find module '@vercel\commerce-local\next.config'** Require stack:
> 
[...]

PR #728, opened 2022-04-06
Import config with POSIX path

markuslewin:
There seems to be an issue in Node.js when importing modules declared in the `exports` field of a `package.json` using Windows paths.

Importing a configuration with `@vercel/commerce-local/next-config` works, but `@vercel\commerce-local\next-config` (as returned by `path.join` on Windows systems) does not.

One possible workaround is to use `path.posix.join` to make sure that the imported path is specified as a POSIX path.

This PR should solve #681.
The `exports` problem might be related to https://github.com/nodejs/node/issues/30169.

### DIFF
--- a/packages/commerce/src/config.cjs
+++ b/packages/commerce/src/config.cjs
@@ -16,7 +16,7 @@ function withCommerceConfig(nextConfig = {}) {
     )
   }
 
-  const commerceNextConfig = importCwd(path.join(provider, 'next.config'))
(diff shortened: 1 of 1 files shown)


dominiksipowicz (approved):

lfades (approved):

dominiksipowicz:
@markuslewin thank you for this contribution. Amazing! 🪟🛫🚀

PR #734, opened 2022-04-19
chore: add prettier vscode settings

dominiksipowicz:
this will enable prettier on save in VS code.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,18 @@
+{
+  "editor.formatOnSave": true,
+  "editor.rulers": [100],
+  "eslint.validate": [
+    "javascript",
+    "javascriptreact",
+    "typescript",
+    "typescriptreact"
+  ],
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,18 @@
...
+    "typescript",
+    "typescriptreact"
+  ],
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "eslint.alwaysShowStatus": true,
+  "eslint.format.enable": true,
+  "prettier.singleQuote": true,
+  "prettier.tabWidth": 2,
+  "prettier.semi": false,
+  "prettier.trailingComma": "es5",
```
lfades, line 16: If we remove this will it default to our `.prettierrc`?

dominiksipowicz: yes, I removed pretty much all and simplify it.

PR #736, opened 2022-04-20
Update Shopify Storefront API Version

laskydev:

PR #737, opened 2022-04-20
Update const.ts

laskydev:


dominiksipowicz:
@laskydev this looks to me like a duplicate of #736 . Can we close this PR?

PR #748, opened 2022-05-08
Swell fixes

samuelfgs:
- Fixed "use-add-item" hook: removed the checkoutId from the request variables. We can't pass the checkoutId to swell because they already use the current session to determine the checkoutId (see use-update-item and use-remove-item).
```
 {
    "code": "permission_error",
    "message": "You are not allowed to update `items.checkoutId`",
    "param": "items.checkoutId"
  }
```

- Fixed search sort: remove the underscore from the sort values ("price_asc", "price_desc"). Swell expect the format `${field} ${direction}`.

- Fixed "use-login" hook: 
```
Unhandled Runtime Error. TypeError: Cannot read properties of undefined (reading '0').
````
Receiving this error when an account doesn't exist.

- Moved the product variants normalization to normalizeProduct: For my use case, I had to expand the "use-search" hook to add product variants too. The normalize function didn't work because it was expecting the variants to be normalized before (see the "get-product operations"). It makes more sense to normalize everything inside the normalize function.

- Add includeSubCategories to "use-search" hook: In swell, you can define a category tree structure. For example
```
Category A
	Category B
		Product B1
		Product B2
Category C
	Product C1
	Category D
		Product D1
		Product D2
```
In the search page, if you filter by Category D, you would get product D1 and D2. But, if you filter C, you would only get product C1, or even worse, if you filter by Category A, the response is "No product found!"

### DIFF
--- a/packages/swell/src/api/operations/get-product.ts
+++ b/packages/swell/src/api/operations/get-product.ts
@@ -20,10 +20,6 @@ export default function getProductOperation({
 
     const product = await config.fetch('products', 'get', [variables.slug])
 
-    if (product && product.variants) {
-      product.variants = product.variants?.results
-    }
-
     return {
       product: product ? normalizeProduct(product) : null,
     }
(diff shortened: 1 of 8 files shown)


okbel:
Amazing! Thank you. Taking a look 👀

Spoutnik97 (approved):

leerob:
Hey there! Thank you for your contribution. We have decided to take Next.js Commerce in a new direction and will be closing out current PRs and issues due to this change. Please see this PR for more details: https://github.com/vercel/commerce/pull/966

PR #761, opened 2022-05-29
fix: put issue templates inside .github folder

SincerelyFaust:
As of right now, we're not able to use the issue forms to submit core bug reports, provider bug reports, feature requests and docs requests as they're not located inside the .github folder, this PR fixes exactly that.
The config is also not working since it isn't located inside the .github folder, therefore it allows creation of blank issues contrary to what it is set up for.


GuiBibeau (approved):
This is a great patch! Thank you very much!

PR #808, opened 2022-08-09
Add availableForSale to variant

powerwlsl:
`Add to cart` button should be disabled when there's no stock for each variant.

```
          <Button
            aria-label="Add to Cart"
            type="button"
            className={s.button}
            onClick={addToCart}
            loading={loading}
            disabled={variant?.availableForSale === false}
          >
            {variant?.availableForSale === false
              ? 'Not Available'
              : 'Add To Cart'}
          </Button>
```

image
image
image


manovotny:
We appreciate the time you took to make this PR. We have decided to take Next.js Commerce in a new direction, but I am happy to say that this feature will be present in the upcoming release. Please see this PR for more details: #966

PR #814, opened 2022-08-10
Add review feature and disable it for BC stores

powerwlsl:
Currently reviews and ratings are hard-coded and we want to disable it for the BC stores. Let me know if there are more steps when adding a feature!

### DIFF
--- a/packages/bigcommerce/src/commerce.config.json
+++ b/packages/bigcommerce/src/commerce.config.json
@@ -2,6 +2,7 @@
   "provider": "bigcommerce",
   "features": {
     "wishlist": true,
-    "customerAuth": true
+    "customerAuth": true,
+    "review": false
   }
 }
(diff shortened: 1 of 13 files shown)


cond0r:
@powerwlsl The site example is more to show the functionality of the commerce package & providers. So in forks, anyone can change the site how is desired; no need to enable or disable it like this.

powerwlsl:
Hi @cond0r! I understand what you meant and I think that would also work.
But we came up with this idea because we wanted to make a default store for BC users less confusing by removing non functioning components. But if the Vercel team doesn't like our approach, we can follow as you said.

leerob:
Hey there! Thank you for your contribution. We have decided to take Next.js Commerce in a new direction and will be closing out current PRs and issues due to this change. Please see this PR for more details: https://github.com/vercel/commerce/pull/966

P.S. we would love to see a BigCommerce version of the new v2 and happy to chat more about this 🙏

PR #826, opened 2022-09-08
Update Shopify URL actual release

laskydev:
The storefront api has changes about metafields that now work differently.

The current shopify storefront api documentation handles the examples with the new version.


cond0r:
Closed in favor of #833 required schema updates & fix deprecations

Issue #512, opened 2021-09-30
Make our core types not extendable


lfades:
You can read more about why we want to do this here: https://github.com/vercel/commerce/issues/429

All our core types include a TS generic that has the base types that might be used by the schema, once our other tasks in https://github.com/vercel/commerce/issues/429 are done we would like to remove the usage of generics from our core types.

E.g the `AddItemHook` would change from:

```ts
export type AddItemHook<T extends CartTypes = CartTypes> = {
  data: T['cart']
  input?: T['itemBody']
  fetcherInput: T['itemBody']
  body: { item: T['itemBody'] }
  actionInput: T['itemBody']
}
```

to:

```tsx
export type AddItemHook = {
  data: Category
  input?: CartItemBody
  fetcherInput: CartItemBody
  body: { item: CartItemBody }
  actionInput: CartItemBody
}
```

We can eventually find a way of converting our types into `interfaces`, but that might cause a few issues at the moment due to the `Provider` type being defined by a generic too and TS complains about it. We can try that one later.

Issue #511, opened 2021-09-30
Update core site types


lfades:
Our site types currently are `Category` and `Brand`, which are defined here: [framework/commerce/types/site.ts](https://github.com/vercel/commerce/blob/main/framework/commerce/types/site.ts), the `brand` type is currently set to `any` and that has to change, currently defined fields are also missing docs.

> We're not exactly sure of how should we continue working with general site types, any recommendations here on how we can scale those types to work with any commerce provider is very welcome!

Every field added needs to have docs, we do that using comments, e.g:

```tsx
/**
 * The checkout's id.
 */
id: string
```

Issue #509, opened 2021-09-30
Define Wishlist schema


lfades:
Our Wishlist type is currently set to `any` and is defined here: [framework/commerce/types/wishlist.ts](https://github.com/vercel/commerce/blob/main/framework/commerce/types/wishlist.ts). Similar to our other core types, we need to have it defined to a shape that can work with any e-commerce provider we throw at it.

Every field defined needs to have docs, we do that using comments, e.g:

```tsx
/**
 * The wishlist's id.
 */
id: string
```

Issue #506, opened 2021-09-29
Update product schema


lfades:
The product schema defined in [framework/commerce/types/product.ts](https://github.com/vercel/commerce/blob/main/framework/commerce/types/product.ts) doesn't have any docs yet, also any updates on fields that might be missing for current or future features are welcome!

Every field defined needs to have docs, we do that using comments, e.g:

```ts
/**
 * The product's id.
 */
id: string
```

PR #831, opened 2022-09-21
Update types

cond0r:
- [x] #504
- [x] #505
- [x] #506
- [x] #509
- [x] #511
- [x] Define customer schema
- [x] #512


packages/commerce/src/types/checkout.ts
```diff
@@ -1,57 +1,97 @@
...
+}
 
-export type CheckoutTypes = {
+export interface CheckoutTypes {
   card?: Card | CardFields
   address?: Address | AddressFields
   checkout?: Checkout
   hasPayment?: boolean
   hasShipping?: boolean
 }
```
lfades, line 56: What is `CheckoutTypes` needed for?

cond0r: I forgot to delete this, it was used before at the generics.

`export type SubmitCheckoutHook<T extends CheckoutTypes = CheckoutTypes> = {`

packages/commerce/src/api/endpoints/cart.ts
```diff
@@ -1,60 +1,62 @@
-import type { CartSchema } from '../../types/cart'
+import type { CartSchema } from '@vercel/commerce/types/cart'
```
lfades, line 1: I know this works but can we keep these imports as relative 🙏 

Ideally avoid imports to the package from the same package for now.

packages/commerce/src/api/utils/errors.ts
```diff
@@ -20,3 +23,31 @@ export class CommerceNetworkError extends Error {
     this.name = 'CommerceNetworkError'
   }
 }
+
+export const getErrorMessage = (error: unknown) => {
```
lfades, line 27: I would prefer to replace this with an error class that's very clear about what it does, like `new CommerceAPINetworkError()`

packages/commerce/src/api/utils/with-schema-parser.ts
```diff
@@ -0,0 +1,54 @@
...
+import { getOperationError } from './errors'
+import { pageSchema } from '../../schemas/page'
+import { siteInfoSchema } from '../../schemas/site'
+import { productSchema, productsPathsSchema } from '../../schemas/product'
+
+export const withSchemaParser =
+  (
+    operation: AllowedOperations,
+    fn: (...args: any[]) => Promise<OperationsData>
+  ) =>
```
lfades, line 14: Could this method receive a type that represents the name of the operation and that has to extend `AllowedOperations`, and then the return type `Promise<OperationsData>` could be based on that name instead?

packages/commerce/src/types/cart.ts
```diff
@@ -1,177 +1,244 @@
...
-  availableForSale?: boolean
-  // The variant's weight. If a weight was not explicitly specified on the
-  // variant this will be the product's weight.
-  weight?: Measurement
-  // The variant's height. If a height was not explicitly specified on the
-  // variant, this will be the product's height.
-  height?: Measurement
-  // The variant's width. If a width was not explicitly specified on the
-  // variant, this will be the product's width.
-  width?: Measurement
```
lfades, line 53: What happened to these fields?

lfades (changes requested):
Added some comments and everything else looks good to me

packages/commerce/src/api/utils/errors.ts
cond0r: This is used & will be used in all the try-catch of all API handlers. Now all the handlers have the error messages repeated. So I created this method because the error from catch comes as unknown, and is used only to send the text message to the response like below. This is already improved in the next PR, for the dynamic API routes. The error class will be thrown in the handlers, then in this method, we will filter what can be shown or not to the client as text messages. Now, the validation error from BigCommerce auth sends back to the res.json ‘An unexpected error occurred, so later we will have a validation error, that we send the actual message to the client, anything else we log the error and send a generic text message as unexpected error.

  ```
try {
    const body = req.body ?? {}
    return await handlers['login']({ ...ctx, body })
  } catch (error) {
    console.error(error)
    const message = getErrorMessage(error)
    res.status(500).json({ data: null, errors: [{ message }] })
  }

```

packages/commerce/src/api/endpoints/cart.ts
cond0r: Yes, this was not intended. This happened when I replaced the types path in all providers to import just from commerce, so I replaced it with find & replace, and matched this one too unfortunately.

packages/commerce/src/api/utils/errors.ts
cond0r: Reverted change, this will reaper on the next PR with dynamic API routes

packages/commerce/src/api/utils/with-schema-parser.ts
cond0r: Changed this to with-operation-callback.ts. In this method, we can chain and do other parallel async transformations to the output. For example, something I have done in another branch to add automatic blurDataURL to images.

packages/commerce/src/types/cart.ts
cond0r: I put them back

lfades (dismissed):
💯

lfades (approved):

PR #870, opened 2022-11-08
Update to Next.js 13

cond0r:
Update the site example to Next.js 13, use the new `next/link` behavior, and the new `next/image`.


packages/shopify/src/utils/get-search-variables.ts
```diff
@@ -11,7 +11,7 @@ export const getSearchVariables = ({
   let query = ''
 
   if (search) {
-    query += `product_type:"${search}" OR title:"${search}" OR tag:"${search}" `
+    query += `(product_type:${search}) OR (title:${search}) OR (tag:${search}) `
```
leerob, line 14: Is this needed?

site/assets/base.css
```diff
@@ -119,3 +119,9 @@ a {
     opacity: 1;
   }
 }
+
+@media not all and (min-resolution: 0.001dpcm) {
+  img[loading='lazy'] {
+    clip-path: inset(0.5px);
```
leerob, line 125: Could you expand on this?

site/components/cart/CartItem/CartItem.tsx
```diff
@@ -86,29 +86,25 @@ const CartItem = ({
...
-              />
-            </a>
+            <Image
+              onClick={() => closeSidebarIfPresent()}
+              className={s.productImage}
+              width={150}
+              height={150}
+              src={item.variant.image?.url || placeholderImg}
+              alt={item.variant.image?.alt || 'Product Image'}
+              unoptimized
```
leerob, line 96: I wonder if we should remove `unoptimized` here... hmm.

site/components/product/ProductCard/ProductCard.module.css
```diff
@@ -69,16 +69,11 @@
 }
 
 .imageContainer {
-  @apply flex items-center justify-center overflow-hidden;
-}
-
-.imageContainer > div {
```
leerob, line 75: This was changed because there's no wrapper now, correct?

site/components/wishlist/WishlistCard/WishlistCard.tsx
```diff
@@ -1,7 +1,7 @@
 import { FC, useState } from 'react'
 import cn from 'clsx'
 import Link from 'next/link'
-import Image from 'next/image'
+import Image from 'next/legacy/image'
```
leerob, line 4: Should this be legacy?

site/package.json
```diff
@@ -35,7 +35,7 @@
     "keen-slider": "^6.7.0",
     "lodash.random": "^3.2.0",
     "lodash.throttle": "^4.1.1",
-    "next": "^12.3.0",
+    "next": "^13.0.0",
```
leerob, line 38: Let's push to latest patch 😄

leerob (changes requested):

leerob:
Thank you so much for doing this!

site/components/product/ProductCard/ProductCard.module.css
cond0r: yes, there was also a div wrapper removed from the component.

site/components/cart/CartItem/CartItem.tsx
cond0r: It makes sense to be optimized here too. Changed to the real size, and also removed the zoom. It works with some demo PNGs here, but with other images looks cropped.

site/assets/base.css
cond0r: This I took from the [docs](https://nextjs.org/docs/api-reference/next/image#known-browser-bugs), can you explain more ?

site/components/cart/CartItem/CartItem.module.css
```diff
@@ -17,13 +17,7 @@
...
 
 .productImage {
-  position: absolute;
-  transform: scale(1.9);
-  width: 100%;
-  height: 100%;
-  left: 30% !important;
-  top: 30% !important;
-  z-index: 1;
+  @appy w-full w-full object-cover;
```
alexdor, line 20: Isn't this a typo?

```suggestion
  @apply w-full w-full object-cover;
```

site/components/common/UserNav/UserNav.tsx
```diff
@@ -59,7 +59,7 @@ const UserNav: React.FC<{
         )}
         {process.env.COMMERCE_WISHLIST_ENABLED && (
           <li className={s.item}>
-            <Link href="/wishlist">
+            <Link href="/wishlist" legacyBehavior>
```
goncy, line 62: Is this one no migrated for some reason?

site/components/search.tsx
```diff
@@ -125,6 +125,7 @@ export default function Search({ categories, brands }: SearchPropsType) {
                     >
                       <Link
                         href={{ pathname: getCategoryPath('', brand), query }}
+                        legacyBehavior
```
goncy, line 128: I image these are because of the onclick needed on the a tag?

goncy (approved):

site/components/search.tsx
cond0r: Yes, because there was a warning if you put onClick on the <Link ...>

lfades (approved):
LGTM

Some things to note for the future:
- When replacing styles with uses of `@apply`, it can be better to remove the use of css for that case and add the classes directly to the element.
- Removing `legacyBehavior` to links is better for the site as this is a starter.

cond0r:
@leerob Could you please review the changes since it blocks the merge?

leerob (approved):

PR #873, opened 2022-11-09
Fixes & Update product image and price based on the variant

cond0r:
- update product image & price on the variant change. Created a product context and updated Shopify and BigCommerce queries and normalizers. [Example here](https://commerce-shopify-k49cnpg87-vercel-solutions-vtest314.vercel.app/product/new-short-sleeve-t-shirt).
- disable add to cart when the selected option is not available (before, the first variant was always auto-selected)
- fixes a case when you visit product 1 that has color & size options, and it was autoselected, when visiting product 2 that had only color, broke variant search, and could not find any variant.
- swatch sizes can have longer strings, add flexible width, and wrap elements
- display product images slider instantly, not on the mount
- will improve the following product tags' line-height
CleanShot 2022-11-09 at 13 14 58@2x
- will fix the following search page title alignment 
CleanShot 2022-11-09 at 13 14 09@2x


gnacoding:
Hi @cond0r 
First thank you very much for your hard work!
I'm currently testing this PR and I have a question: could you please point me to the file/line which is responsible for the image to switch when selecting a variant?

cond0r:
The image changes by checking the selected variant provided by the Product context. So on the line below, it searches for the image index in the images thumbs with the same URL as the image for that variant. So for that to work, the provider needs to set the Image property on the variant in the first place. This PR currently automatically sets that for Shopify & BigCommerce. What provider do you use ?

https://github.com/vercel/commerce/blob/7748d885cc95b3ae925514903a94d38939db419f/site/components/product/ProductSlider/ProductSlider.tsx#L48

site/components/product/ProductSlider/ProductSlider.tsx
```diff
@@ -109,6 +124,8 @@ const ProductSlider: React.FC<ProductSliderProps> = ({
                 ...child,
                 props: {
                   ...child.props,
+                  width: 132,
```
goncy, line 127: Are these values meant to be fixed?

site/components/product/Swatch/Swatch.module.css
```diff
@@ -1,9 +1,9 @@
 .swatch {
   box-sizing: border-box;
   composes: root from '@components/ui/Button/Button.module.css';
-  @apply h-10 w-10 bg-primary text-primary rounded-full mr-3 inline-flex
+  @apply h-10 w-10 bg-primary text-primary rounded-full inline-flex
   items-center justify-center cursor-pointer transition duration-150 ease-in-out
-  p-0 shadow-none border-accent-3 border box-border select-none;
+  p-0 shadow-none border-accent-3 border box-border select-none !mr-0;
```
goncy, line 6: What is `!mr-0`? First time seeing it 👀

goncy (dismissed):

site/components/product/Swatch/Swatch.module.css
cond0r: The ! before any tailwind class is like setting  !important, swatch had a margin-right (that I thought comes from the button that composes, but it was following line. So I removed the margin-right and added a gap to the parent because when the swatch wrapped and fell underneath, it didn't have a margin-top.

PR #880, opened 2022-11-17
Fix 500 when a product is not found

### LINKED ISSUE #879: Product route returns 500 when product is not found
### Describe the feature you'd like to request

Since the addition of `with-operation-callback.ts` when a product is not found the route returns a 500 error.

For example https://demo.vercel.store/product/not-a-product-slug

An Error is thrown at https://github.com/vercel/commerce/blob/6c2610584de973f90777defc0f5daa94330b5335/packages/commerce/src/api/utils/with-operation-callback.ts#L17 and it comes from a failed validation at: https://github.com/vercel/commerce/blob/6c2610584de973f90777defc0f5daa94330b5335/packages/commerce/src/api/utils/with-operation-callback.ts#L24

The problem here seems to be that when a product is not found `data.product` is undefined

### Describe the solution you'd like

When a product is not found the route should return a 404

### Describe alternatives you've considered

I've considered to wrap al the code in `site/pages/product/[slug].ts` in a try catch but it's cleaner to change https://github.com/vercel/commerce/blob/6c2610584de973f90777defc0f5daa94330b5335/packages/commerce/src/api/utils/with-operation-callback.ts#L24 to 
```ts
productSchema.nullable().parse(data.product ?? null)
```

slackerzz:
fix #879 

I'm using this with BigCommerce

### DIFF
--- a/packages/commerce/src/api/utils/with-operation-callback.ts
+++ b/packages/commerce/src/api/utils/with-operation-callback.ts
@@ -21,7 +21,7 @@ export const withOperationCallback =
 const parse = ({ name, data }: Operation) => {
   switch (name) {
     case 'getProduct':
-      productSchema.nullable().parse(data.product)
+      productSchema.optional().parse(data.product)
       break
     case 'getAllProducts':
       z.array(productSchema).parse(data.products)
--- a/site/pages/product/[slug].tsx
+++ b/site/pages/product/[slug].tsx
@@ -34,7 +34,9 @@ export async function getStaticProps({
   const { products: relatedProducts } = await allProductsPromise
 
   if (!product) {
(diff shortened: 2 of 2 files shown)


packages/commerce/src/api/utils/with-operation-callback.ts
```diff
@@ -21,7 +21,7 @@ export const withOperationCallback =
 const parse = ({ name, data }: Operation) => {
   switch (name) {
     case 'getProduct':
-      productSchema.nullable().parse(data.product)
+      productSchema.nullable().parse(data.product ?? null)
```
cond0r, line 24: `.nullable()` can be replaced with `.nullish()` so it accepts `null or undefined`, then there is no need for ` ?? null`

cond0r (approved):

PR #886, opened 2022-11-22
Improve error handling

cond0r:
- improves Zod parsing error message when there is no path/field
- catch Saleor's throttle errors
CleanShot 2022-11-22 at 10 44 37@2x


leerob (approved):
Code seems fine but the saleor build is failing - related?

cond0r:
> Code seems fine but the saleor build is failing - related?

I created this PR to catch the error before it was not catching it and returned data undefined. Saleor fails because, like in the screenshot above, the environment is limited and can't handle that many requests. I talked with the Saleor team, and they will increase that limit.

PR #889, opened 2022-11-25
Fix BigCommerce login

### LINKED ISSUE #888: Login not working with BigCommerce
### Describe the feature you'd like to request

After [this](https://github.com/vercel/commerce/commit/c75b0fc0012b2c31c40c6b1584b1a72c0dbeea7f) commit the login stopped working with BigCommerce.

The problem seems to be in https://github.com/vercel/commerce/blob/main/packages/bigcommerce/src/api/endpoints/login/login.ts where a new Response is returned and not the one manipulated in https://github.com/vercel/commerce/blob/main/packages/bigcommerce/src/api/operations/login.ts with the `Set-Cookie` header.

After a signup the `SHOP_TOKEN` cookie is correctly set in the client, but every call to `/api/commerce/customer` return 404, so I believe that the changes at
https://github.com/vercel/commerce/blob/main/packages/bigcommerce/src/api/endpoints/customer/get-logged-in-customer.ts#L33-L39
introduced a regression

### Describe the solution you'd like

To be able to login

### Describe alternatives you've considered

none

slackerzz:
~~Partially~~ fix  #888

### DIFF
--- a/packages/bigcommerce/src/api/endpoints/customer/get-logged-in-customer.ts
+++ b/packages/bigcommerce/src/api/endpoints/customer/get-logged-in-customer.ts
@@ -34,7 +34,7 @@ const getLoggedInCustomer: CustomerEndpoint['handlers']['getLoggedInCustomer'] =
         getLoggedInCustomerQuery,
         undefined,
         {
-          'Set-Cookie': `${config.customerCookie}=${token}`,
+          cookie: `${config.customerCookie}=${token}`,
         }
       )
       const { customer } = data
--- a/packages/bigcommerce/src/api/operations/login.ts
+++ b/packages/bigcommerce/src/api/operations/login.ts
@@ -67,7 +67,7 @@ export default function loginOperation({
       const newCookie = concatHeader(prevCookie, cookie)
 
       if (newCookie) {
-        res.headers.set(
+        response.headers.set(
           'Set-Cookie',
           String(Array.isArray(newCookie) ? newCookie.join(',') : newCookie)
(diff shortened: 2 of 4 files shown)


slackerzz:
@cond0r now the problem is in [packages/commerce/src/api/utils/index.ts](https://github.com/vercel/commerce/blob/main/packages/commerce/src/api/endpoints/customer/index.ts#L22) where the validation against the customerSchema fails

cond0r:
> @cond0r now the problem is in [packages/commerce/src/api/utils/index.ts](https://github.com/vercel/commerce/blob/main/packages/commerce/src/api/endpoints/customer/index.ts#L22) where the validation against the customerSchema fails

Yes, I will investigate more and provide a fix by next week. From what I remember, login & Signup had some unresolved issues from the past. Also, I saw during my quick tests that the validations from BigCommerce don't appear and show generic messages. That will be covered in my fix PR also,  Maybe this should be closed since it isn't the complete fix for it.

slackerzz:
@cond0r changing the `customerSchema` to:

```
export const customerSchema = z.object({
  customer: z.object({
    id: z.string(),
    firstName: z.string(),
    lastName: z.string(),
    email: z.string().optional(),
    phone: z.string().optional(),
    company: z.string().optional(),
    notes: z.string().optional(),
    acceptsMarketing: z.boolean().optional(),
  })
})
```

seems to fix this issue

cond0r:
Thank you! this will be addressed in #918

PR #894, opened 2022-11-29
Footer theme switcher

cond0r:
CleanShot 2022-11-29 at 20 09 18@2x


site/components/common/Footer/Footer.tsx
```diff
@@ -9,6 +10,13 @@ import { Logo, Container } from '@components/ui'
 import { I18nWidget } from '@components/common'
 import s from './Footer.module.css'
 
+const ThemeSwitcher = dynamic(() => import('@components/ui/ThemeSwitcher'), {
+  ssr: false,
+  loading: () => (
+    <div className="w-24 h-10 rounded-md bg-accent-2 animate-pulse" />
```
lfades, line 16: This loading state is smaller than the actual input so there's a CLS. For this case I don't see a reason to use lazily load the component, it's small and doesn't load large libraries.

lfades (changes requested):
Added a comment, also I think it makes more sense to have a 90 degree arrow when the switchers are closed, instead of an upwards arrow, e.g:
image
So then it will switch between 90 (closed) and 180 degrees (opened)
image

site/components/common/Footer/Footer.tsx
cond0r: I will make the rotation and remove the lazy loading. I did that because the themes array and the current theme come from the client side, so it showed a hydration issue instead of fixing it, showing nothing, decided to show a placeholder until loads and is read.

site/components/ui/ThemeSwitcher/ThemeSwitcher.tsx
```diff
@@ -1,18 +1,23 @@
...
 
 const ThemeSwitcher = () => {
+  const [mounted, setMounted] = useState(false)
   const [display, setDisplay] = useState(false)
   const { theme, themes, setTheme } = useTheme()
 
+  useEffect(() => setMounted(true), [])
+
+  if (!mounted) return null
+
```
lfades, line 17: This feels wrong, something like this makes more sense to me: https://github.com/pacocoursey/next-themes/issues/153#issuecomment-1324246080

lfades (changes requested):
Left another comment, and also the spacing of the arrows doesn't match properly:
image
The lang arrow has less left margin than the theme.

lfades (approved):
Very nice!

PR #908, opened 2022-12-07
Custom fields

cond0r:
This PR implements custom fields & metafields to the product. 

1. The new customFields product attribute represents an array of key-value strings from BigCommerce provider (others still evaluating); add it to the product query, types & normalizer. And a component that shows the values in the `ProductSidebar`.
2. Metafields represent more advanced custom fields that are grouped by namespace. Shopify & BigCommerce has it (others evaluating). Implemented the types and queries for Shopify. 

For Shopify, we need to provide the following parameters at the query; you need to send the identifiers which means namespace + key that you want to fetch. I added that as an optional parameter to the `getProduct` operation. So if you want to get (those namespace & keys, you set yourself in the dashboard), for example:
CleanShot 2022-12-07 at 12 03 46@2x

And you access it like this. Instead of an array and then finding the value by a key, I propose to have it like this. The metafields attribute is a record with the key to be the namespace, and the value is a child record with the key and the rest of the fields as value. 
CleanShot 2022-12-07 at 13 58 48@2x

At the Shopify normalizer, I automatically transformed the value, which is often a string that can be parsed as JSON, into an HTML string for booleans, dimensions, measuring units & dates (internationalized based on the current locale), multi-lines, and lists/arrays of values just mentioned. These transformed values can be accessed from the `valueHtml`, and if the raw value is needed, you can access the value attr, which is the value sent by Provider.
￼
CleanShot 2022-12-07 at 12 06 01@2x


goncy (dismissed):

packages/bigcommerce/package.json
```diff
@@ -63,6 +63,7 @@
     "react-dom": "^18"
   },
   "devDependencies": {
+    "@manifoldco/swagger-to-ts": "^2.1.0",
```
lfades, line 66: Where is this being used?

packages/commerce/src/types/product.ts
```diff
@@ -117,6 +117,33 @@ export interface Product {
...
+   * Advanced custom fields that can be added to a product. They are used to store additional information about the product, in a structured format, grouped by namespaces.
+   * @example
+   * {
+   *  // Namespace, the container for a set of metadata
+   *  reviews: {
+   *   // Key of the metafield
+   *   rating: {
+   *    key: 'rating',
+   *    value: 4,
+   *    valueHtml: '&#9733;&#9733;&#9733;&#9733;&#9734;',
```
lfades, line 140: This needs another comment to know what's this HTML about.

packages/shopify/src/types/metafields.ts
```diff
@@ -0,0 +1,39 @@
...
+  | 'list.page_reference'
+  | 'list.product_reference'
+  | 'list.rating'
+  | 'list.single_line_text_field'
+  | 'list.url'
+  | 'list.variant_reference'
+  | 'list.volume'
+  | 'list.weight'
+
+export type MetafieldType = LiteralUnion<MetafieldTypes>
```
lfades, line 39: What does this type do? I tried it out locally and using this type accepts any string.

packages/shopify/src/utils/metafields.ts
```diff
@@ -0,0 +1,118 @@
...
+    case 'json':
+    case 'single_line_text_field':
+    case 'product_reference':
+    case 'page_reference':
+    case 'variant_reference':
+    case 'file_reference':
+    default:
+      return value
+  }
+}
```
lfades, line 108: If we're doing some custom logic over the metafields, would it be possible to only use one value instead of having both `value` and `valueHtml`?

lfades (changes requested):
Added some comments

packages/shopify/src/utils/metafields.ts
cond0r: I left the value to be the original provider value if the user wants to do his parsing to be able to have the original value. Same we have `description` & `descriptionHtml`. At Shopify, that value most times is a JSON stringified.

packages/bigcommerce/package.json
cond0r: I need to double-check, but it was a missing peer dependency from graphql codegen from what I remember.

packages/shopify/src/utils/metafields.ts
lfades: What do you think are good use cases for this? because I think it's better to have transformers as utils for cases like this.

cond0r: For example, the case was to have the auto-transformed value of the custom field. Let's take the weight as an example. The value that comes from Shopify is `{unit: MILLIMETERS, value: 10}`, and it wanted to provide the auto-transformed value `10mm` (or a link, rating, date, etc.), but if the user wants the original value for his transformation, to can access it, and not modify the provider for that. Yes agree with utils, but it is better to transform right after fetch and normalization, then call that util from the site folder, then conditionally call & include that since not all providers will have that

packages/shopify/src/types/metafields.ts
cond0r: Yes, since extends `string` and we use utils like `.split('.')` later on. It was not used to constraint specific "types", it was more used for autocomplete in the switch.

packages/shopify/src/utils/metafields.ts
cond0r: Initially, I wanted to provide a component, parse, and conditionally show there. Since that would be used just by Shopify and create confusion, I decided on parsing on normalizer. Let me know which direction should we decide.

PR #918, opened 2022-12-14
Fix auth & wishlist

cond0r:
- fix auth & register after a regression was introduced by the Dynamic API PR, unable to test since auth was disabled on preview deployments.
- show the actual validation responses thrown by BigCommerce
- fix wishlist API endpoints/operations schema & normalization after Type updates PR
- fix customer schema validation


packages/bigcommerce/src/api/endpoints/signup/signup.ts
```diff
@@ -22,28 +21,27 @@ const signup: SignupEndpoint['handlers']['signup'] = async ({
...
+      const message = Object.values(error.data.errors).join('<br />')
+      if (message) {
+        throw new CommerceAPIError(message, {
           status: 400,
-          code: 'duplicated_email',
+          code: 'invalid_request',
         })
       }
-    } else {
-      throw error
```
lfades, line 36: I understand improving the error handling here, but why was the duplicated email check removed?

packages/bigcommerce/src/api/operations/get-customer-wishlist.ts
```diff
@@ -41,18 +39,22 @@ export default function getCustomerWishlistOperation({
...
 
     const wishlist = data[0]
 
     if (includeProducts && wishlist?.items?.length) {
-      const ids = wishlist.items
-        ?.map((item) => (item?.productId ? String(item?.productId) : null))
-        .filter((id): id is string => !!id)
+      const ids = []
+
+      for (let i = 0; i < wishlist.items.length; i++) {
```
lfades, line 51: This is better as a `for of`

packages/bigcommerce/src/api/operations/login.ts
```diff
@@ -63,19 +63,13 @@ export default function loginOperation({
         cookie = cookie.replace(/; SameSite=none/gi, '; SameSite=lax')
       }
 
-      const prevCookie = response.headers.get('Set-Cookie')
-      const newCookie = concatHeader(prevCookie, cookie)
```
lfades, line 67: The import for `concatHeader` was not removed, and that util might not be needed anymore if it's no longer used here, also, why is this part changing?

packages/bigcommerce/src/api/utils/fetch-graphql-api.ts
```diff
@@ -7,7 +7,7 @@ const fetchGraphqlApi: (getConfig: () => BigcommerceConfig) => GraphQLFetcher =
   async (
     query: string,
     { variables, preview } = {},
-    options: { headers?: HeadersInit } = {}
+    headers?: HeadersInit
```
lfades, line 10: I don't see this change as a good idea, having an options object that includes headers is far better IMO.

packages/commerce/src/api/utils/index.ts
```diff
@@ -44,7 +44,9 @@ export const transformRequest = (req: NextApiRequest, path: string) => {
     body = JSON.stringify(req.body)
   }
 
-  return new NextRequest(`https://${req.headers.host}/api/commerce/${path}`, {
+  const url = new URL(req.url || '/', `https://${req.headers.host}`)
+
+  return new NextRequest(url, {
```
lfades, line 49: The `path` parameter of this function is no longer used after this change, why was this changed? not mentioned in the description 🤔

packages/commerce/src/schemas/customer.ts
```diff
@@ -5,14 +5,16 @@ export const getCustomerAddressBodySchema = z.object({
...
 export const customerSchema = z.object({
-  id: z.string(),
-  firstName: z.string(),
-  lastName: z.string(),
-  email: z.string().optional(),
-  phone: z.string().optional(),
-  company: z.string().optional(),
-  notes: z.string().optional(),
-  acceptsMarketing: z.boolean().optional(),
+  customer: z.object({
```
lfades, line 8: Why is it needed to add a nested `customer` object in the customer schema?

lfades (changes requested):
Added multiple comments.

packages/bigcommerce/src/api/endpoints/signup/signup.ts
cond0r: This displays the actual error and if there are multiple errors. And I checked, and the BigCommerce message was more explicit. I modified this because BigComerce might throw many invalid requests/validations that, at signup, showed it a generic 500 error message for a password that was seven characters but not a numbers & letters combination. It still catches the duplicated email.

packages/commerce/src/schemas/customer.ts
cond0r: This I took from another PR that reported that error, but yes, will modify it in the endpoint handler and will make there z.object({customer: customerSchema}), because the operation, returns that as customer obj.

packages/commerce/src/api/utils/index.ts
cond0r: `req.url` comes with full path + query search, which was lacking before.

packages/bigcommerce/src/api/operations/login.ts
cond0r: Since the res is not passed after dynamic handler PR, I removed that from here too.

packages/bigcommerce/src/api/utils/fetch-graphql-api.ts
cond0r: This was aligned with the changes already **merged**, so back then you missed this change on the previous PR. I reverted these changes to reflect your comment with all providers to have headers passed in an options obj.

PR #966, opened 2023-04-18
Next.js Commerce refresh.

leerob:
We're making some updates to Next.js Commerce. Everything prior to this PR marks what we're calling [`v1`](https://github.com/vercel/commerce/releases/tag/v1) as a point in time to be able to reference and still use going into the future. The current architecture of Commerce is a multi-vendor, interoperable solution, including:

- [Shopify](https://shopify.vercel.store/)
- [Swell](https://swell.vercel.store/)
- [BigCommerce](https://bigcommerce.vercel.store/)
- [Vendure](https://vendure.vercel.store/)
- [Saleor](https://saleor.vercel.store/)
- [Ordercloud](https://ordercloud.vercel.store/)
- [Spree](https://spree.vercel.store/)
- [Kibo Commerce](https://kibocommerce.vercel.store/)
- [Commerce.js](https://commercejs.vercel.store/)
- [SalesForce Cloud Commerce](https://salesforce-cloud-commerce.vercel.store/)

All features can be toggled on or off, and it's easy to change between commerce providers. To support this, we needed to create a ["commerce metaframework"](https://github.com/vercel/commerce/blob/d1d9e8c4343e70d72df6e00ec95f6ea4efbb7c54/packages/commerce/new-provider.md) where providers could confirm to an API spec to add support for Next.js Commerce. While this worked and was successful for `v1`, we have different design goals and ambitions for `v2`.

**What You Need To Know**

- `v1` will not be updated moving forward. If you need to reference `v1`, you will still be able to clone and deploy the version tagged at this release.
- `v2` will be shifting to be a single provider vs. provider agnostic. Other providers are welcome to fork this repository and swap out the underlying `lib/` implementation that connects to the selected commerce provider (Shopify). This architecture was chosen to reduce the surface area of the codebase, remove the intermediate metaframework layer for provider-interoperability, and enable usage with the latest Next.js and React features.
- We will be sharing more about `v2` in the future as we continue to iterate before the marked release.

Thank you so much! 🎉


michaelbromley:
Thanks for providing some clarity on the status of the project. Over at Vendure we have been getting quite a lot of support questions about the status of this repo. It has been clear for a while that v1 had hit a bit of a dead end.

I think the decision you took makes sense - being platform agnostic is _theoretically_ good but in practice any one project will only ever use a single backend, so all the abstraction required to make it agnostic is immediately just unwanted bloat that over-complicates the architecture. I think this is also what Vue Storefront also eventually learned.

There's still going to be demand for Next.js in the Vendure community of course, but we might go with building our own starter kit tailored to our APIs, like we did with our existing Remix, Qwik & Angular starters. But I'll be keeping an eye on this repo still to see to what degree it makes sense to recommend to our users to fork and adapt it for their Vendure projects.

megetron:
Thank you for the update. I must admit that I am not as fond of this decision. I appreciate the performance and architecture of the frontend Next.js Commerce, as well as the ability to choose the backend. Removing this option means I am now forced to use the Shopify solution, which is not the one I currently use.

However, I am hopeful that the "lib" folder's architecture will allow me to replace the GraphQL queries with those I already use on v1. This would make replacing the "lib" a simple copy-and-paste process with some code replacements.

I look forward to your future updates regarding the v2 version.

manovotny:
We understand that the new direction will come with mixed feelings. @michaelbromley is right. Being platform agnostic is theoretically good, but typically comes with a more complex architecture to allow the flexibility. That was evident with v1. As you can see, this PR removed close to 150k lines of code while only adding 6k.

The goals of the new template are to:

* Provide a singular focused template (we chose Shopify) to clearly and concisely demonstrate how to create a best in class ecommerce experience.
* Demonstrate Next.js’s latest features, such as App Router, RSC (React Server Components), Suspense, Route Handlers, Edge functions, new fetching and caching paradigms, Metadata, dynamic OG images, and more. 
* Include missing real-world ecommerce features such as out of stocks, order history, cross variant / option availability (aka. Amazon style), dynamically driving all content and features via the provider (ie. collections, menus, pages, etc.), and more.

As for supporting other providers, that’s where [Vercel’s Template Marketplace](https://vercel.com/templates) comes in. We recommend and welcome templates that will produce the same goals listed above.

app/search/page.tsx
```diff
@@ -0,0 +1,41 @@
+import SearchResults from 'components/layout/search/results';
+import { defaultSort, sorting } from 'lib/constants';
+import { getProducts } from 'lib/shopify';
+
+export const runtime = 'edge';
```
m1nsuplee, line 5: Thank you for providing an amazing demo.👍

I have a question to ask you. 

Could you please explain the rationale behind choosing to use the Edge runtime in this page?

uchitwai:
c5a5fb739d086e5de6d3a0077211af8ee3f86c6e

PR #975, opened 2023-04-20
Recommended product square update

StephDietz:
Updated the UI of the recommended products to match the tiles on the search pages


manovotny (approved):

components/layout/product-grid-items.tsx
```diff
@@ -0,0 +1,29 @@
+import Grid from 'components/grid';
+import { GridTileImage } from 'components/grid/tile';
+import { Product } from 'lib/shopify/types';
+import Link from 'next/link';
+
+export default async function ProductGridItems({ products }: { products: Product[] }) {
```
leerob, line 6: Should this be an async function?

manovotny: Uhh... No. There is no reason for it to be async.

manovotny: Addressed in #977.

PR #982, opened 2023-04-21
Adds sitemap.

manovotny:
https://commerce-shopify-git-sitemap-vercel-solutions-vtest314.vercel.app/sitemap.xml

CleanShot 2023-04-21 at 15 50 47@2x

### DIFF
--- a/app/sitemap.ts
+++ b/app/sitemap.ts
@@ -1,0 +1,33 @@
+import { getCollections, getPages, getProducts } from 'lib/shopify';
+import { MetadataRoute } from 'next';
+
+const baseUrl = process.env.NEXT_PUBLIC_VERCEL_URL
+  ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
+  : 'http://localhost:3000';
+
+export default async function sitemap(): Promise<Promise<Promise<MetadataRoute.Sitemap>>> {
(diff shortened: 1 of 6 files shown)


app/sitemap.ts
```diff
@@ -0,0 +1,33 @@
+import { getCollections, getPages, getProducts } from 'lib/shopify';
+import { MetadataRoute } from 'next';
+
+const baseUrl = process.env.NEXT_PUBLIC_VERCEL_URL
+  ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
```
manovotny, line 5: Can this be done dynamically like this or should we hardcode it to `demo.vercel.store`?

app/sitemap.ts
```diff
@@ -0,0 +1,33 @@
...
+  : 'http://localhost:3000';
+
+export default async function sitemap(): Promise<Promise<Promise<MetadataRoute.Sitemap>>> {
```
leerob, line 8: Promise-ception

app/sitemap.ts
leerob: This should work, vs. hardcoding. There isn't a `metabaseBase` here similar to the built-in SEO support AFAIK, but could be wrong.

leerob (approved):

PR #997, opened 2023-04-26
Add Vercel Web Analytics

tobiaslins:
Start capturing custom events for `Search` `Add to cart` `Checkout`


leerob:
Could you address the merge conflicts? Will this work when you clone and deploy or do you need to set something up ahead of time?

tobiaslins:
@leerob I think @timolins mentioned this on Slack. The only problem is that people will need to enable Web Analytics on the dashboard, else they will see a `404` for the script.
I'm wondering if we could introduce a way to automatically enable it when someone deploys with `@vercel/analytics` installed and not manually disabled it yet.

thomasbritton:
Interesting feature add, but I feel the nature of requiring this to be enabled in the Vercel Project Settings etc kind of makes me feel that shouldn't be added to this particular codebase as a out of the box solution.

And if you are on the Hobby plan, you can only have the Web Analytics on 1 project if you have multiple projects.

It almost feels like it should be used a specific template/example such as 'with-web-analytics', similar to https://github.com/vercel/next.js/tree/canary/examples

leerob:
We have decided to leave this out of the template, but should be able to be easily added if the customer wants to do so 🙏

PR #1019, opened 2023-05-12
Fixes search page bugs.

### LINKED ISSUE #1017: Full-text query is lost when a sort criteria is selected
How to reproduce the bug :
- In the landing page, fill the search input with "rainbow" and hit enter.
- The search page displays 2 items matching the query.
- Select "Sort by price : low to high".
- The query is lost, the search page shows all the items.

I understand the value of having an explict & shareable url in the search page.
But that leads to a shared store, read with the `useSearchParams` hook, and updated with `<Link .. >` components.

Considering the sorts, filters and queries as the context of the search page (of the app ?), I'd suggest to 
- use the React Context API to store them
- build a custom `useSearchCriteria` to deal with the state stored in the context.

If it's empty, the context could be initialized with the search params. 
A "share this list" button may implement a shareable url generation.

manovotny:
* Retains sorting when searching and navigating categories
* Removes search term when navigating categories

Fixes #1017

### DIFF
--- a/app/search/[collection]/page.tsx
+++ b/app/search/[collection]/page.tsx
@@ -4,6 +4,7 @@ import { notFound } from 'next/navigation';
 
 import Grid from 'components/grid';
 import ProductGridItems from 'components/layout/product-grid-items';
+import { defaultSort, sorting } from 'lib/constants';
 
 export const runtime = 'edge';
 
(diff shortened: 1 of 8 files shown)


leerob (approved):
Should we also clear the search input?

manovotny:
> Should we also clear the search input?

That's what I meant by "Removes search term when navigating categories". Are you seeing something different?

PR #1022, opened 2023-05-16
Created a bash script to automate the process of .env file creation and also GITPODIFIED the repo.

KapadiaNaitik:
Bash script automates the process of .env file creation and with the help of it I created a gitpod config for this repo so developers don't have to go through the hassle of manual configuration and can make changes on the cloud ide on the go.


leerob:
Hey! Thank you for the suggestion, but we'd prefer not to add a bash script for this.

Issue #1025, opened 2023-05-19
/product/${product.handle} not resolving for linked instagram/facebook stores 


Seyaji:
The SEO and product link for shopify products is by default `/products/*item*` 
This template uses `/product/*item*`
This is not an issue when routing traffic internally on the vercel hosted site but when linking from an external integration (like instagram) the link will return a 404. 

I did not find a way to change the default product URL on Shopify , although there may be a way.

This is an issue when trying to handle multiple sales channels that feed the website.

leerob:
https://github.com/vercel/commerce/pull/1026#issuecomment-1556291798

PR #1026, opened 2023-05-19
update /product URL to /products

Seyaji:
#1025 
update /product URL to /products to support meta commerce integration…s and shopify SEO


leerob:
Can you explain more why this is needed? It seems like a cosmetic thing, but I might be missing something.

leerob:
Okay, nevermind. I see the linked issue.

> linking from an external integration (like instagram) the link will return a 404.

You should not be linking to the Shopify managed domain name, but the domain name of your headless store where you control the URL structure. This could be `/items/*` if you wanted.

Seyaji:
If you try and link your Shopify store to Instagram using this template it will gather the products but when you click the link on Instagram it will fail, as the expected URL is /products. This means that linking an Instagram shop as a sales channel will cause users to receive a 404 when being referred from Instagram. This is quite a harsh limitation.

Seyaji:
@leerob I am linking to my headless store. 
It is not possible to customise the item url as these are generated by the meta commerce integration.

leerob:
I am not following – you must be doing something differently than what is in this repository. In this repo, you can set up your URLs in whichever pattern you want. It's using Shopify as the headless commerce provider. You could also change this out, if you want. You have complete control.

Seyaji:
Thanks for taking the time to discuss this, 

The issue is that meta commerce draws the catalog from shopify and generates the links automatically.

For example, you could have `/category/item` as the URL pattern on the headless store, but when the instagram shop is populated from the shopify catalog the links all generate with `/products/item`. 

So in this case the **"View on Website"** link for the product on instagram will be `/products/item` and return a 404. 

While its possible to change the URL of the item on shopify I haven't found a way to rename the `/products` endpoint or to change the way the links generate, or even edit these links at all, its not possible through instagram or meta commerce manager. 

So it seems the only way to get the integration to work is to maintain the shopify URL pattern.

PR #1050, opened 2023-06-16
Replaces Route Handlers with Server Actions

manovotny:


manovotny:
This is ready for review. I noticed an existing issue with the cart that I was going to try and also fix in this PR, but decided to not to, otherwise it would get too noisy with unrelated changes. 

The issue, which is not due to these changes, it logged in https://github.com/vercel/commerce/issues/1051.

leerob (approved):

leerob:
I noticed the same bug, good call out 👍

PR #1054, opened 2023-06-21
Fixes bundle size for Hobby plan users

### LINKED ISSUE #1047: The Edge Function "[page]/opengraph-image" size is 1 MB and your plan size limit is 1 MB.
Not able to deploy on a pristine e-commerce clone.
image

manovotny:
This is working now with the latest updates to Next.js. You can see that [this is working on my fork](https://github.com/manovotny/commerce/pull/2), which uses a Hobby plan, and it's now deploying without error. We are now well under the 1MB limit, as seen below.

### Before

CleanShot 2023-06-22 at 10 07 29@2x

### After 

CleanShot 2023-07-05 at 09 10 31@2x

This also updates all dependencies and remove removes un-used dependencies.

Fixes #1047


leerob:
Do we need this in here all the time or should this be temporary so you can identify the issue?

manovotny:
@leerob I can go either way.... Once I diagnose, I can remove it if you prefer.

manovotny:
Now using these PR to work with the Next.js team to reduce bundle size. I have changed the PR to a draft until we feel like we have a path forward.

manovotny:
This is working now with the latest updates to Next.js. You can see that [this is working on my fork](https://github.com/manovotny/commerce/pull/2), which uses a Hobby plan, and it's now deploying without error.

leerob (approved):
Love it, thank you!

PR #1057, opened 2023-06-25
feat(typesafety):return-types|readonly

admirsaheta:
Utility function | Additional Typesafety
 -  Explicit return type ( string ) -> The return type should be a string since the function returns a URL.
 -  Implicit readonly on types.
 - Template literals with conditional operator instead of string concatenation


leerob:
This is just stylistic right?

leerob:
Closing as stylistic, thank you regardless 🙏

PR #1066, opened 2023-07-10
Replace Icons with heroicons

StephDietz:


components/layout/search/filter/dropdown.tsx
```diff
@@ -45,7 +45,7 @@ export default function FilterItemDropdown({ list }: { list: ListItem[] }) {
         className="flex w-full items-center justify-between rounded border border-black/30 px-4 py-2 text-sm dark:border-white/30"
       >
         <div>{active}</div>
-        <Caret className="h-4 rotate-90" />
+        <ChevronRightIcon className="h-4" />
```
manovotny, line 48: This is pointing the incorrect direction. 

CleanShot 2023-07-10 at 15 13 02@2x

manovotny (changes requested):
One minor fix, otherwise LGTM!

PR #1067, opened 2023-07-10
New header UI

StephDietz:


manovotny (changes requested):

manovotny (changes requested):
There's still a couple of break points that produce awkward site name shifts / stacking. 

We probably just need to either: 

1. Go to the mobile header much sooner.
2. Move search over next to the cart.

CleanShot 2023-07-11 at 13 20 39@2x

CleanShot 2023-07-11 at 13 20 52@2x

evilrabbit:
I would go to mobile sooner or we can hide `ACME STORE` and keep the icon on smaller screens.

manovotny (approved):

PR #1070, opened 2023-07-11
Product page UI refresh

StephDietz:


app/product/[handle]/page.tsx
```diff
@@ -76,34 +74,25 @@ export default async function ProductPage({ params }: { params: { handle: string
   };
 
   return (
-    <div>
+    <div className="px-4">
```
Comment (manovotny) [inline app/product/[handle]/page.tsx:77]: Can this be changed to a fragment and the style moved to the `div` below the script?

app/product/[handle]/page.tsx
```diff
@@ -2,16 +2,14 @@ import type { Metadata } from 'next';
 import { notFound } from 'next/navigation';
 import { Suspense } from 'react';
 
-import Grid from 'components/grid';
+import { GridTileImage } from 'components/grid/tile';
 import Footer from 'components/layout/footer';
-import ProductGridItems from 'components/layout/product-grid-items';
```
Comment (manovotny) [inline app/product/[handle]/page.tsx:7]: Is this component used anywhere else? Can it be deleted?

components/cart/add-to-cart.tsx
```diff
@@ -50,15 +50,16 @@ export function AddToCart({
...
       className={clsx(
-        'flex w-full items-center justify-center bg-black p-4 text-sm uppercase tracking-wide text-white opacity-90 hover:opacity-100 dark:bg-white dark:text-black',
+        'relative flex w-full items-center justify-center rounded-full bg-blue-600 p-4 tracking-wide text-white hover:opacity-90',
         {
           'cursor-not-allowed opacity-60': !availableForSale,
           'cursor-not-allowed': isPending
         }
       )}
     >
+      <div className="absolute left-0 ml-4">{!isPending ? <PlusIcon className="h-5" /> : null}</div>
```
manovotny, line 60: What if, instead of the `PlusIcon` and `LoadingDots` having separate spaces, they occupy the same space? So... 

```
<div className="absolute left-0 ml-4">{isPending ? <LoadingDots className="bg-white" /> : <PlusIcon className="h-5" />}</div>
```

manovotny (changes requested):
Last comment, and IDK where it's happening in the code exactly...

I kinda dislike how the gallery images get bigger or smaller depending on how many there are. I think it would be best to have them consistently the same size. I actually prefer the smaller ones in the screenshot below. 

CleanShot 2023-07-12 at 14 26 22@2x

I still like this location over what is in Figma, but Figma did have these as circles instead of squares. Do you have a preference, @StephDietz?

app/product/[handle]/page.tsx
Comment (StephDietz) [reply app/product/[handle]/page.tsx:7]: We use it in the search page

Comment (manovotny) [reply app/product/[handle]/page.tsx:7]: 👍

StephDietz:
> 

Agreed, I like them small better. Didn't notice this was happening bc i didn't see any items that had that many images. I set the images to have a fixed width instead and made them on the smaller side

manovotny (changes requested):
Love the changes. One more thing I just noticed. 🤦‍♂️ 

Can we center the images under the primary image? Looks odd on larger screens when we don't. 

CleanShot 2023-07-12 at 15 02 18@2x

StephDietz:
> Love the changes. One more thing I just noticed. 🤦‍♂️
> 
> Can we center the images under the primary image? Looks odd on larger screens when we don't.
> 
> CleanShot 2023-07-12 at 15 02 18@2x

Done!

manovotny (approved):
Bueno!

PR #1071, opened 2023-07-12
Separate cart-icon into open-cart and close-cart

StephDietz:


components/cart/open-cart.tsx
```diff
@@ -1,24 +1,18 @@
-import { ShoppingCartIcon, XMarkIcon } from '@heroicons/react/24/outline';
+import { ShoppingCartIcon } from '@heroicons/react/24/outline';
```
manovotny, line 1: Can you rename the exported component to `OpenCart` please?

Sorry I can't comment on the actual line as it didn't change. 🫤

manovotny (changes requested):
Crazy small changes, but I like this move over all. Feels right.

manovotny (approved):

PR #1072, opened 2023-07-13
UI refresh for search page

StephDietz:


components/grid/tile.tsx
```diff
@@ -7,11 +7,13 @@ export function GridTileImage({
   active,
   labelPosition,
   labels,
+  hoverBorder,
```
manovotny, line 10: Stepping back and looking at the site as a whole... Do you think it's odd that search is the only page we have hover? I'm thinking we should add hover to all instances by default. This would cover the carousel, related products, and search results, right? If we do add it to all of these, we should also add it to the tiles on the homepage too. Thoughts?

manovotny:
Home page carousel?

manovotny (approved):

PR #1099, opened 2023-07-24
Adds dynamic product label font size

manovotny:
Uses [CSS container queries](https://www.smashingmagazine.com/2021/05/complete-guide-css-container-queries/) (which has [decent brower support](https://caniuse.com/css-container-queries)) to dynamically change the product title and price font size and display of currency code, enabled via the [tailwindcss-container-queries]() plugin.

The video below will demonstrate how this works on the search page. You'll notice the font size and showing of `USD` changing based on the size of the grid item.

https://github.com/vercel/commerce/assets/446260/c5b4d1f1-af4e-4ee3-8fab-eaa0f09ad71f

Also refactored the `GridTileImage` / `Label` properties:

* Removed `isSmall` as it's not longer necessary.
* Moved `position` into the `label` prop object.

### DIFF
--- a/app/product/[handle]/page.tsx
+++ b/app/product/[handle]/page.tsx
@@ -125,8 +125,7 @@ async function RelatedProducts({ id }: { id: string }) {
             >
               <GridTileImage
                 alt={product.title}
-                labels={{
-                  isSmall: true,
+                label={{
                   title: product.title,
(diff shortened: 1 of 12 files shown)


leerob (approved):
I've been meaning to dig into these... this is awesome. The future of CSS is 🤯

PR #1124, opened 2023-07-28
Remove unnecessary `useEffect`s in search filter

kevin940726:
I was just reading the code for inspiration and found this piece of code confusing. I believe it's an example of unnecessary `useEffect`s as described in the [React docs](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state). I've not tested this though as I don't have a paid Shopify account to set this up locally 😅. Feel free to close this if I'm missing something obvious! 🙇 

A bonus task would be to check if we could make these RSCs now that we don't need states or effects.


manovotny (approved):
@kevin940726 yep, I believe you're right. Nice catch! It looks like there may be opportunity to turn it into an RSC, but I'm going to merge this as a good first step. Thanks for the contribution!

PR #1134, opened 2023-08-02
Adds better sitemap error handling

### LINKED ISSUE #1120: Provide a better error message for `sitemap.xml` failures
In continuation of #1010, we should wrap the creation of the sitemap in a try/catch so we can output a better message when environment variables are missing or incorrectly entered.

manovotny:
### Before

CleanShot 2023-08-01 at 20 23 43@2x

### After

CleanShot 2023-08-01 at 20 22 03@2x

Fixes #1120

### DIFF
--- a/app/sitemap.ts
+++ b/app/sitemap.ts
@@ -1,11 +1,16 @@
 import { getCollections, getPages, getProducts } from 'lib/shopify';
 import { MetadataRoute } from 'next';
 
+type Route = {
+  url: string;
+  lastModified: string;
+};
+
 const baseUrl = process.env.NEXT_PUBLIC_VERCEL_URL
   ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
   : 'http://localhost:3000';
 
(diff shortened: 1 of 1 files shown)


app/sitemap.ts
```diff
@@ -1,11 +1,16 @@
...
+type Route = {
+  url: string;
+  lastModified: string;
+};
+
 const baseUrl = process.env.NEXT_PUBLIC_VERCEL_URL
   ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
   : 'http://localhost:3000';
 
-export default async function sitemap(): Promise<Promise<Promise<MetadataRoute.Sitemap>>> {
```
leerob, line 8: `PromisePromisePromisePromise`

leerob (approved):

app/sitemap.ts
manovotny: At least it was keeping all those promises it was making. 😂

PR #1143, opened 2023-08-03
Removes priority on third grid item on home page

manovotny:


styfle (approved):
I think this will improve mobile scores, not sure about desktop but certainly the first image is considered LCP

leerob (approved):

Issue #1123, opened 2023-07-27
Search page loading state does not reappear after initial search


TonsiTT:
Searching for something in the search bar will take you to the /search route and show the loading state.
However, further searches from the /search route do not show the loading state again.  It is more of an issue. on slower internet devices where it would seem as if the webpage is stuck since there is no indicator otherwise to show it is fetching data in the background and getting new results

Any way to force the loading state again after the search parameter in the URL changes?

manovotny:
I have Chrome throttled to 100 kbit/s in this video. I ***think*** I see what you're talking about here. You'll see when I switch from `shirt` to `hat`, the titles on the cards change nearly instantly, but the pics take a few seconds to reload with the right images. 

Is this what you're seeing or something different?

https://github.com/vercel/commerce/assets/446260/9ec02872-c1b3-4dc5-a751-813d3e954451

TonsiTT:
When you initially go to `shirt`, there is a loading skeleton, as seen in the photo. However, going from `shirt` to `hat`, it doesn't reappear. If fixed, it would solve the issue of reloading with the correct images.
IMO, a skeleton should reappear every time the search param changes like `shirt` to `hat` 
 
![image](https://github.com/vercel/commerce/assets/50301560/b5257fe6-d501-4e1f-81fc-79bf2ffe11b5)

leerob:
We'll likely move to this pattern: https://buildui.com/posts/instant-search-params-with-react-server-components

PR #1152, opened 2023-08-05
issue#1123 fix

sellimenes:
Issue #1123 , added Suspense to show loading skeleton on every single search request.

### DIFF
--- a/app/search/page.tsx
+++ b/app/search/page.tsx
@@ -2,6 +2,7 @@ import Grid from 'components/grid';
 import ProductGridItems from 'components/layout/product-grid-items';
 import { defaultSort, sorting } from 'lib/constants';
 import { getProducts } from 'lib/shopify';
+import { Suspense } from 'react';
 
 export const runtime = 'edge';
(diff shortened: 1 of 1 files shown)


leerob:
I'm not sure if this is necessary, it already feels really fast when going between searches to me. What do you see?

sellimenes:
> I'm not sure if this is necessary, it already feels really fast when going between searches to me. What do you see?

When I try on slow internet, it doesn't feel juicy enough I think. With Suspense, it just 'feels' better

muco-rolle:
I have a similar issue on my project, where it takes 1 to 2 seconds before the search parameter is added to the address bar.

sellimenes:
> I have a similar issue on my project, where it takes 1 to 2 seconds before the search parameter is added to the address bar.

Can you also please try to use Suspense and share your experiences?

leerob:
We'll likely move to this pattern: https://buildui.com/posts/instant-search-params-with-react-server-components

PR #1156, opened 2023-08-08
Update README.md

BrocksiNet:
Adding Shopware Template and Demo


manovotny:
@BrocksiNet thanks for extending and adding Shopware to the list. A couple of things to note:

1. We're trying to keep provider alphabetized. Can you adjust Shopware in the list.
2. We noticed some changes to the theme (header and search page), which is fine. Just know that the maintenance burden on your end might be a bit heavier when you pull in upstream updates, which I am sure you've thought of. Just saying it outloud. 😊

BrocksiNet:
@manovotny The order is now alphabetic. And yes we added some more features like pagination and so on. Sure it will be a bit harder but I hope we can handle this. :joy:

manovotny (approved):

PR #1165, opened 2023-08-09
Added Umbraco provider

mattbrailsford:
Adding a link to the Umbraco + Umbraco Commerce implementation of the Vercel Commerce demo.

Worth noting, our repo has a little different structure as we wanted to include the back end project as well so we have a `backend/main` and a `frontend/main` branch accordingly. The `frontend/main` branch can easily be merged with upstream changes from the vercel commerce repo.


leerob:
For some reason, the images seem slow to load here.

https://github.com/vercel/commerce/assets/9113740/f3128549-8634-48ad-848c-706d1e723932

leerob:
CleanShot 2023-08-09 at 18 13 27@2x
Checkout gave me a 500.

mattbrailsford:
> For some reason, the images seem slow to load here.
> 
>  CleanShot.2023-08-09.at.18.11.20.mp4

Hmmm, should these be cached via Next? I wonder if it’s because the images were already processed. I copied them from Shopify. Is there anyway I can get a copy of the originals?

mattbrailsford:
500 error has been resolved. If you can provide the raw product images, I'll update those and see if that affects image load times. Otherwise I'm not 100%.

PR #1193, opened 2023-09-07
Single product page price updated

kurucaner:
Problem:
On the individual product pages, the displayed price doesn't reflect the true price of the selected variant.

Reproduction:
Navigate to: [ACME Drawstring Bag (Black, Size: 9 x 12 inch)](https://demo.vercel.store/product/acme-drawstring-bag?color=Black&size=9+x+12+inch). The price shown is $12, yet when added to the cart, the actual cost is $9.

Solution:
This PR aims to fix this discrepancy by dynamically updating the displayed price based on the chosen product variant.

### DIFF
--- a/components/cart/actions.ts
+++ b/components/cart/actions.ts
@@ -1,9 +1,11 @@
 'use server';
 
+import { TAGS } from 'lib/constants';
 import { addToCart, createCart, getCart, removeFromCart, updateCart } from 'lib/shopify';
+import { revalidateTag } from 'next/cache';
 import { cookies } from 'next/headers';
 
-export const addItem = async (variantId: string | undefined): Promise<String | undefined> => {
+export async function addItem(prevState: any, selectedVariantId: string | undefined) {
   let cartId = cookies().get('cartId')?.value;
   let cart;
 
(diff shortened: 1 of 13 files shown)


components/product/product-description.tsx
```diff
@@ -1,31 +1,51 @@
+'use client'
```
BC-krasnoshapka, line 1: I think changing whole `product-description` to client component has negative consequences.
Alternatively can we transfer state between `VariantSelector` and `AddToCart` via URL params and `useSearchParams`?

kurucaner: Thank you for the heads-up. I've removed 'use client' in the product-description. Instead, I've introduced a useSelectedVariant hook which is now being used in both the Price and AddToCart components. I hope this aligns with the proposed approach.

MurusMe:
Thanks, @kurucaner you're a wizard! It really helped me out. RetroFlynn.com, the changes you made really worked for me. Thanks!

BC-krasnoshapka (approved):

krzywiecki:
What happened to this PR. Prices for specifi variants still don't seem to work.

Issue #1197, opened 2023-09-15
Issue while building the site from vercel while deploying the project


chandanitme:
![image](https://github.com/vercel/commerce/assets/49269952/d83928dd-0520-4958-a50d-4bc3f4b80e4a).

manovotny:
This is because the `SHOPIFY_STORE_DOMAIN` environment variable isn't set. You can read more about it in our [Shopify integration guide](https://vercel.com/docs/integrations/shopify#configure-environment-variables).

chandanitme:
@manovotny 
below are the environment variables I am passing.
COMPANY_NAME="Vercel Inc."
TWITTER_CREATOR="@vercel"
TWITTER_SITE="https://nextjs.org/commerce"
SITE_NAME="Next.js Commerce"
SHOPIFY_REVALIDATION_SECRET=""
SHOPIFY_STOREFRONT_ACCESS_TOKEN=""
SHOPIFY_STORE_DOMAIN="[ordercloudchandan].myshopify.com"
Please correct me if anything wrong.Please let me know how to get the below values.

SHOPIFY_REVALIDATION_SECRET=""
SHOPIFY_STOREFRONT_ACCESS_TOKEN=""

manovotny:
@chandanitme remove the brackets from `SHOPIFY_STORE_DOMAIN`. The value should be `ordercloudchandan.myshopify.com`.

chandanitme:
I did that but still the issue comes

Thanks. Chandan

chandanitme:
@manovotny .I have made the change but still the error.
![image](https://github.com/vercel/commerce/assets/49269952/2a7b78ad-ec0f-4ee2-b30b-2e600e462222)

COMPANY_NAME="Vercel Inc."
TWITTER_CREATOR="@vercel"
TWITTER_SITE="https://nextjs.org/commerce"
SITE_NAME="Next.js Commerce"
SHOPIFY_REVALIDATION_SECRET=""
SHOPIFY_STOREFRONT_ACCESS_TOKEN=""
SHOPIFY_STORE_DOMAIN="ordercloudchandan.myshopify.com"

manovotny:
@chandanitme then it's because `SHOPIFY_STOREFRONT_ACCESS_TOKEN` is missing. As I mentioned before, the [Shopify integration guide](https://vercel.com/docs/integrations/shopify#configure-environment-variables) instructs you how to get this in your Shopify admin.

chandanitme:
[...]

Issue #1189, opened 2023-09-04
Issue when trying to deploy this repo locally: TypeError [ERR_INVALID_URL]: Invalid URL, message: 'URL is malformed "/api/2023-01/graphql.json"


borilshishkov:
I have cloned the repo and trying to deploy it locally while following all instructions. Using environment variables defined in .env.example, 
after that I installed Vercel CLI and linked Vercel with Github -` vercel link` and generated `.vercel` folder.
In the end after running 
```
pnpm install
pnpm dev
```

I encounter this issue:

```
fedora commerce]$ sudo pnpm dev

> @ dev /home/dev/Documents/dev/vercel/commerce
> next dev

- info Loaded env from /home/dev/Documents/dev/vercel/commerce/.env.local
- warn You have enabled experimental feature (serverActions) in next.config.js.
- warn Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use at your own risk.

- ready started server on 0.0.0.0:3000, url: http://localhost:3000
- event compiled client and server successfully in 268 ms (20 modules)
- wait compiling...
- event compiled client and server successfully in 289 ms (20 modules)
- info Loaded env from /home/dev/Documents/dev/vercel/commerce/.env.local
- info Loaded env from /home/dev/Documents/dev/vercel/commerce/.env.local
- wait compiling /page (client and server)...
- event compiled client and server successfully in 8.8s (1923 modules)
- error {
  cause: 'TypeError [ERR_INVALID_URL]: Invalid URL',
  status: 500,
  message: 'URL is malformed "/api/2023-01/graphql.json". Please use only absolute URLs - https://nextjs.org/docs/messages/middleware-relative-urls',
  query: '\n  query getMenu($handle: String!) {\n    menu(handle: $handle) {\n      items {\n        title\n        url\n      }\n    }\n  }\n'
}
- error {
  cause: 'TypeError [ERR_INVALID_URL]: Invalid URL',
  status: 500,
  message: 'URL is malformed "/api/2023-01/graphql.json". Please use only absolute URLs - https://nextjs.org/docs/messages/middleware-relative-urls',
[...]

PR #1198, opened 2023-09-15
Adds environment variable validation

manovotny:
We get 1 -3 issues or comments a week about build errors (see #1189 and #1197, for example), which are due to missing environment variables.

This PR add some validation during the build process to ensure absolutely required env vars are present and throws a clearer error message.

CleanShot 2023-09-16 at 19 59 28@2x


fucha-co:
i've also seen multiple matters of confusion around these Shopify variables — including my own — and upon requesting clarification the responses have been very dismissive — including to my own. it's clearly a point of confusion for many so perhaps the relevant docs aren't quite clear enough and/or could be better

manovotny:
@fucha-co happy to make changes! Do you have suggestions on what should be made more clear or how it should be made more clear?

PR #1209, opened 2023-10-01
Improve form submissions, update packages.

leerob:
- Switch to `<form>` and `useFormState` for progressively enhanced forms
- Fix some accessibility issues
  - Improved color contrast
  - Added live region for announcing errors
  - Switch from button `disabled` to `aria-disabled`
- Update all dependencies
- Ditch `router.refresh()` to `revalidateTag` and tag the cart `fetch`
- Add a key to `input` to that changing search values correct re-renders and clears value
- Updated Prettier config to work with latest ESM versions (+ Tailwind config)
- Prevent duplicate clicks on buttons causing multiple form submissions


components/cart/add-to-cart.tsx
```diff
@@ -5,8 +5,67 @@ import clsx from 'clsx';
...
+  return (
+    <button
+      onClick={(e: React.FormEvent<HTMLButtonElement>) => {
+        if (pending) e.preventDefault();
+      }}
+      aria-label="Add to cart"
+      aria-disabled={pending}
+      className={clsx(buttonClasses, {
+        'hover:opacity-90': true,
+        'cursor-not-allowed': pending
```
manovotny, line 59: Should we just use `disabledClasses` here?

pnpm-lock.yaml
```diff
@@ -1,18 +1,22 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
```
manovotny, line 4: This is what is causing the errors on CI. Pnpm will set `--frozen-lockfile` when running in CI and with `autoInstallPeers: true`, it will likely install things the lock files doesn't have.

Did you add this on purpose? How did it get added?

manovotny:
Personally, I don't like the `.bind` calls. I don't see the value in using them over hidden inputs, at least in this scenario. I'll stand down if this is the preferred direction, but I wanted to mention it.

pnpm-lock.yaml
leerob: I didn't add this on purpose, hmm. And blowing away the lockfile and re-running keeps this setting.

manovotny: Did you intentionally update the `package.json` pnpm version? I wonder if that is slipping it in. 

https://github.com/vercel/commerce/pull/1209/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R3

leerob: I tried to intentionally update it to fix it, with no effect.

manovotny (approved):

PR #1234, opened 2023-11-06
Remove revalidate

joostmeijles:
Remove revalidate as there is currently no cache support for the Edge runtime.


leerob:
Will be kept here since edge is removed https://github.com/vercel/commerce/pull/1236

PR #1237, opened 2023-11-12
Update tailwind.config.js

james0r:
Added type defs for custom Tailwind plugin args.


leerob:
I'd like to move to the TS file instead 🙏

PR #1241, opened 2023-11-13
Removing useEffects that have no side effect and alter the state.

odin-nido:
It is recommended by the React Docs to avoid unnecessarily changing the state within a useEffect.
As React Docs says: "Storing information from previous renders like this can be hard to understand, but it’s better than updating the same state in an Effect."


leerob:
Thank you for the suggestion. I think this is fine to keep as is, but definitely this is a good option to consider if folks want to optimize here.

PR #1245, opened 2023-11-16
disable the image optimization

megetron:
disabling the next/image optimization because:
1. the headless CMS do the image optimizations instead of next/image.
2. image/next optimization makes links such as `http://127.0.0.1:3001/media-main/image1-100x124.webp&w=3840&q=75`.
this link is illegible and throws 404 since after the .webp you expect to have "?".


megetron:
sorry, missed the previous PR. closing

PR #1257, opened 2023-12-01
Implement 'Compare at Price' Feature

rondo10:
Description

This Pull Request introduces the 'Compare at Price' feature to our e-commerce platform. The primary goal of this feature is to enhance the user shopping experience by displaying a comparison between the original price and the current sale price of products. This feature is especially beneficial for promotional events, allowing customers to see the value they are getting from discounted prices.

How It Works

The feature retrieves the compare-at price and the normal price from the product fetch. These prices are then displayed side by side on the product page. A strikethrough style is applied to the original price to denote the discount.


leerob:
Thanks for the suggestion, but we're not going to add this here. However, you can use it in your fork!

PR #1260, opened 2023-12-15
Correct default cart tax currency

matthew-petrie:
The cart defaults the tax to be `USD` if there is no tax amount set for the cart. This is an unhelpful default due to international and multi currency stores where the currency changes with the cart (most of the this repo handles this correctly).

This changes defaults the currency to the same currency as the cart total currency if there are no taxes set for the current cart.

Before:
Screenshot 2023-12-15 at 14 55 28

After:
Screenshot 2023-12-15 at 14 55 38


leerob (approved):
Thank you!

PR #1289, opened 2024-01-30
Fix #1283: Added hover effect to search icon.

### LINKED ISSUE #1283: add hover to search icon

The-Kaif:
Add hover effect to search icon

Fixes #1283

In this commit, I've added a hover effect to the search icon as requested in issue #1283. Now, users will experience a visual cue when hovering over the search icon, enhancing the overall user experience. 

Contributed by: the-kaif


leerob:
Thank you for the suggestion, but we are okay without this!

PR #1304, opened 2024-02-27
Add Ecwid by Lightspeed provider

meteor-ec:
Hi there! 

This is Ilnur from the Ecwid by Lightspeed team. 

We have added our service as a provider. Please add us to the list.

Thanks a lot!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Vercel is happy to partner and work with any commerce provider to help them get 
 
 - Shopify (this repository)
 - [BigCommerce](https://github.com/bigcommerce/nextjs-commerce) ([Demo](https://next-commerce-v2.vercel.app/))
+- [Ecwid by Lightspeed](https://github.com/Ecwid/ecwid-nextjs-commerce/) ([Demo](https://ecwid-nextjs-commerce.vercel.app/))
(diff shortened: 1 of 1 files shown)


manovotny:
Hey @meteor-ec! This is a good start!

I looked at the [changes you made](https://github.com/vercel/commerce/compare/main...Ecwid:ecwid-nextjs-commerce:main) and I have some questions.

1. The checkout page is very slow and the code for the page is interesting. Is this what you expect users to do? Feels like this could be improved.
2. Why are you adding `<link rel="prefetch" href="/api/revalidate" as="fetch" />` to the `<head>` in the Layout? What are you trying to do or what does this achieve? 
3. `image: product.featuredImage?.url` in `productJsonLd` could produce an invalid schema as it could be undefined. It's be better to change the the model to always return a url, and in `lib/ecwid` return a default images if there isn't a featured image.
4. The changes to CartModal for not displaying on the checkout page could be improved. Rearranging the structure of your Layouts or [using multiple root Layouts](https://nextjs.org/docs/app/building-your-application/routing/route-groups#creating-multiple-root-layouts) would be the better solve here.

manovotny:
@meteor-ec thanks for making those adjustments.

meteor-ec:
Thank you so much @manovotny! :handshake:

PR #1327, opened 2024-04-19
Make variant / image updates use `useOptimistic` and `startTransition`

leerob:
https://github.com/vercel/commerce/assets/9113740/e05ad411-b505-4af9-b46c-6c602f10da77


components/product/variant-selector.tsx
```diff
@@ -21,14 +22,21 @@ export function VariantSelector({
   const router = useRouter();
   const pathname = usePathname();
   const searchParams = useSearchParams();
+  const [optimisticVariants, setOptimsticVariants] = useOptimistic(variants);
```
mhart, line 25: Nit
```suggestion
  const [optimisticVariants, setOptimisticVariants] = useOptimistic(variants);
```

components/product/variant-selector.tsx
```diff
@@ -74,15 +74,36 @@ export function VariantSelector({
...
+                  });
+
+                  optimisticOptions.set(optionNameLowerCase, value);
+
+                  setOptimsticVariants(newOptimisticVariants);
+                  setOptimisticOptions(new URLSearchParams(optimisticOptions.toString()));
+
+                  const optionUrl = createUrl(pathname, optimisticOptions);
+
+                  // Navigate without page reload
```
mhart, line 104: I'm not sure this comment's 100% accurate. It seems that the navigation can't complete without a full page render – at least in the sense that the search params are up-to-date, etc. While the UI is ready to interact with due to the useOptimistic hook, it's still waiting for the full page render before the search params update, so it's now open to race conditions for anything reading from the search params (eg, adding to cart)

leerob:
Replaced by https://github.com/vercel/commerce/pull/1365

nikosantis:
Hi @leerob 
What happens to the optimistic state if the server action fails?

leerob:
It reverts back to the last previous successful state – which is good! You can then show errors like a toast if you want.

PR #1352, opened 2024-07-12
Refactor <Prose> component

chogyejin:
- Remove as assertions for prop that is already inferred to be `string`.
- Change component props(`TextProps`) to inline for code consistency.


leerob (approved):
Thank you!

PR #1363, opened 2024-07-25
feat(design): Show carousel above the fold on desktop

vvo:
Before this commit, we would not see the carousel without scrolling. The top images are so big that they take up most of the space. This made the website look buggy (opinion). Thus, I am proposing this change.

**Before**
CleanShot 2024-07-25 at 11 38 17@2x

**After**
CleanShot 2024-07-25 at 11 39 03@2x

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+.env*.local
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -36,3 +36,4 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+.env*.local
```
vvo, line 39: was added automatically by vc env pull

leerob (approved):

PR #1364, opened 2024-07-25
Optimistic cart

leerob:
Switch to `useOptimistic` for instant feedback on cart +/-. You'll notice that the Server Actions run serially to ensure that all write mutations succeed. If you try to navigate to checkout, it kicks off an action to redirect. If the other mutations have not yet completed, it will be queued behind those and wait for all mutations to settle. This ensures you can't navigate to the cart while cart additions or subtractions are happening.

### DIFF
--- a/components/cart/actions.ts
+++ b/components/cart/actions.ts
@@ -4,6 +4,7 @@ import { TAGS } from 'lib/constants';
 import { addToCart, createCart, getCart, removeFromCart, updateCart } from 'lib/shopify';
 import { revalidateTag } from 'next/cache';
 import { cookies } from 'next/headers';
+import { redirect } from 'next/navigation';
 
 export async function addItem(prevState: any, selectedVariantId: string | undefined) {
   let cartId = cookies().get('cartId')?.value;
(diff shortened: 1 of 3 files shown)


michael-land:
@leerob  It looks like this didn't work as expected when the item is taxable. The optimisticUpdate shows incorrect tax and total

leerob:
I don't believe any of the items in this demo have tax set up - could you share a video?

michael-land:
From this video, you can see that the total and tax values do not match between the calculated value and the API response

https://github.com/user-attachments/assets/2682e9cb-7a81-4ca9-acf6-b43aa6cfa214

leerob:
Oh, okay! Yeah, this isn't implemented here because we aren't using tax on the items. Happy to accept upstream PRs for solutions here. 

https://github.com/vercel/commerce/pull/1365

michael-land:
I don't believe this is possible with the Storefront API, as Shopify does not expose tax rate information. One way to work around this is to display the subtotal instead of the total, and for the tax rate, just say 'calculated at checkout'.

Issue #1272, opened 2024-01-07
Shopify Navigation and menus


tanjunior:
I am following the [Integration Guide](https://vercel.com/docs/integrations/shopify) it didn't exactly show an example for what value to enter for menu items. I tried a few ways and couldn't get it to work.

![image](https://github.com/vercel/commerce/assets/11617693/e9fdd216-f77c-4cdf-9684-49a9019b1880)
When clicked on About in the footer it goes to domain/pages/about instead of domain/about

And if I set it as 
![image](https://github.com/vercel/commerce/assets/11617693/f0d90dd1-945e-497c-8489-4d3aa82771d4)
Routing will only work in accessed from domain root

same as for
![image](https://github.com/vercel/commerce/assets/11617693/fc031832-6a04-4ea8-9315-530e2ad9cc90)

danieldanielecki:
I have the same problem, did you fix it?

simpslandyy:
I'm also experiencing the same issue. :(

danieldanielecki:
Oh, no sorry. I simply can't see navigation at all. 
Have you faced such an issue?

simpslandyy:
Similar, I'm noticing all the pages go to **<domain>/pages/<subpage> instead of <domain>/<subpage>** and idky. :(

issackj1:
One possible solution for the issue of `:domain/pages/:subpage` is to follow the suggestion of OP and change the page links to `/about` and `/frequently-asked-questions` instead of using the page "About". 
![Screenshot 2024-01-18 at 7 14 38 PM](https://github.com/vercel/commerce/assets/25108114/c0f20de6-0574-4865-aee9-df75cd7e95e4)
If the site is hosted on vercel, it might be necessary to clear the cache on vercel to see the changes reflected on the navigation bar. I found that i had to do that before I could see the new navigation to `:domain/:subpage` instead of `:domain/pages/:subpage`.

I still have the last issue that OP mentioned.
[...]

PR #1368, opened 2024-08-09
no-store cache for shopify menu fetch

backbone-link:
Shopify currently does not seem to provide a webhook for updating the Navigation Menus under the Online Store. Also this function was using the tag for collections, which is not related. Removing the tag and adding a `no-store` cache.


backbone-link:
Related comment in the open issues: https://github.com/vercel/commerce/issues/1272#issuecomment-2278858259

PR #1415, opened 2024-12-03
Added Prodigy Commerce as a commerce provider

polykoi:


manovotny:
1. There's some `console.log` statements in the `instrumentation.ts`. Did you want to leave those in there? 
2. Does Prodigy platform not have a concept of non-product pages? Is that why there's fake pages in `lib/prodigy/api/menus.ts`?

polykoi:
1. It is intended to show those messages in logs. We connect the app to the platform automatically in `instrumentation.ts`. So those logs can give some clues if something is going wrong with connection process.
2. No we don't yet have such concept. So we stubbed those pages to conform more to the original template. Users would be able to create and manage non-product pages as part of their app.

manovotny:
1. Sounds good. We'll leave them in.
2. Would it make sense to leave it this way (markup in string) or would it be better to create standard `page.tsx` files? All the markup in strings feels icky.

polykoi:
I didn't like it much too. It was just a quick solution to conform better to existing codebase and simplify syncing maintenance in the future.  
I've moved them to separate pages already.

PR #1455, opened 2025-05-15
Optimize product page route shell

unstubbable:
_stacked on #1454_

Due to the product pages using the `ProductProvider`, which uses `useSearchParams`, the product route shells are currently pretty much empty. This PR attempts to sketch out how we could push the `useSearchParams` further down, and use it more granularly, so that skeletons can be used as suspense fallbacks wherever product selections can be made that influence the search params. Thus allowing us to statically prerender most of the product details into the route shell. The selection elements then become interactive on hydration. There's still some minor CLS that should be fixed, but it's a start.

### DIFF
--- a/app/product/[handle]/page.tsx
+++ b/app/product/[handle]/page.tsx
@@ -4,7 +4,7 @@ import { notFound } from 'next/navigation';
 import { GridTileImage } from 'components/grid/tile';
 import Footer from 'components/layout/footer';
 import { Gallery } from 'components/product/gallery';
-import { ProductProvider } from 'components/product/product-context';
+import { GallerySkeleton } from 'components/product/gallery-skeleton';
 import { ProductDescription } from 'components/product/product-description';
 import { HIDDEN_PRODUCT_TAG } from 'lib/constants';
 import { getProduct, getProductRecommendations } from 'lib/shopify';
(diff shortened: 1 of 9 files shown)


unstubbable:
I fell into a trap here. Because of the absence of `generateStaticParams`, Next.js is considering this route as high-cardinality, and will always use the fallback shell for all routes. It won't produce route shells on-demand when visiting the product pages with specific params. What tripped me up was that with `next start` we currently do generate those route shells. That needs to be fixed in Next.js.

darekrossman:
There's still the issue of `useOptimistic` being in the hook. You wont get optimistic updates across multiple components using optimistic state, like changing the color in variant selector and having the image swap to that color, since every hook has its own state. My solution is to keep the ProductProvider, but only use it to hold selection state. The `useProduct` hook will initialize the provider state from `useSearchParams`, and update the provider state immediately when changing attributes. A simpler approach is to mount a single client component under the Provider, and have it `useSearchParams` and then `useEffect` to set Provider selection state just once on mount. 

The use of useOptimistic here is questionable really, since all we really need to do is set the initial selections from searchParams once on mount. After that, we can drive the UI from the Provider's state and get instant updates across components. We still update the URL when selections change, but only for deep linking.

